PR #759, opened 2014-12-28
Make -proxy set all network types, avoiding a connect leak.

rnicoll:
Previously -proxy was not setting the proxy for IsLimited networks, so
 if you set your configuration to be onlynet=tor you wouldn't get an
 IPv4 proxy set.

The payment protocol gets its proxy configuration from the IPv4 proxy,
 and so it would experience a connection leak.

This addresses issue #5355 and also clears up a cosmetic bug where
 getinfo proxy output shows nothing when onlynet=tor is set.

Conflicts:
	src/init.cpp

Rebased-From: 3c777141349ad82d679a278df0619968af53c23
Github-Issue: #5358

### DIFF
--- a/contrib/debian/examples/bitcoin.conf
+++ b/contrib/debian/examples/bitcoin.conf
@@ -6,7 +6,7 @@
 # Run on the test network instead of the real dogecoin network.
 #testnet=1
 
-# Connect via a socks4 proxy
+# Connect via a SOCKS5 proxy
 #proxy=127.0.0.1:9050
 
 # Use as many addnode= settings as you like to connect to specific peers
--- a/contrib/debian/manpages/bitcoin-qt.1
+++ b/contrib/debian/manpages/bitcoin-qt.1
@@ -32,10 +32,7 @@ Set database cache size in megabytes (default: 25)
 Specify connection timeout in milliseconds (default: 5000)
 .TP
 \fB\-proxy=\fR<ip:port>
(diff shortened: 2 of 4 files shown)


langerhans:
There are a few related commits to this, here especially the one removing SOCKS4 support. I'll try to list the corresponding ones here, might be worth packing them together?

Commit | Issue/PR
----|----
bitcoin/bitcoin@0127a9be14089b3f352ec349b2ecf4488234a4d8 | Issue bitcoin/bitcoin#4327
bitcoin/bitcoin@a339a37b28176135c7fe5adbd3b50da77a45373d | Issue bitcoin/bitcoin#4327
bitcoin/bitcoin@d817187bdbeacb2eb6bd23d7faa38e667770e7e6 | Issue bitcoin/bitcoin#4597
bitcoin/bitcoin@aabe61cb37373913c9df4c86a7eed4e5b0bbf7a1 | Issue bitcoin/bitcoin#5322
bitcoin/bitcoin@59abcefc2eec7ede6f098dbcf7323aef6e746a4a | n/a

rnicoll:
Unfortunately those mostly depend on other commits earlier in the queue, so trying to apply them out of sequence just creates unnecessary headaches. I've applied the documentation change, but the others I think will have to wait.

langerhans:
Okay, well, then I'll merge here. The others should be applied eventually anyways :)

PR #761, opened 2014-12-28
ProcessGetData(): abort if a block file is missing from disk

rnicoll:


langerhans:
bitcoin/bitcoin#4173 also included bitcoin/bitcoin@8c93bf4c2857cdc01f0e5efca2becef08d2c5f1e 

It should be okay to pull it in seperately, if you don't want it here. But I feel a bit of packing things together might be worth it? What do you think?

langerhans:
Looks good

PR #772, opened 2014-12-31
Qt: Add GUI view of peer information. #4133

rnicoll:


langerhans:
This breaks currently in Travis since it tries to build with `--disable-wallet` too. bitcoin/bitcoin@a92aded70ec2346c3f07ff1cf8eb97101a76912f should fix it, but also fixes a similar problem with another pathc. Should keep that in mind. Will merg for now until the other one is pulled and both get the fix commit :)

langerhans:
I like the feature very much tho :)

PR #776, opened 2015-01-02
Updated license date

benhc123:
Updated copyright/license date to 2015 for new year.


langerhans:
Slightly wrong branch, as 1.9-dev is current ;)

I'll just commit this myself later since we have three times the same PR currently in :D

PR #780, opened 2015-01-04
Enable limited TX replacement

rnicoll:
Enables TX replacment in the case where the conflicted TX is not final, and the
replacement is final.

This patch diverges from Bitcoin Core, who do not allow any replacement currently,
for concerns of DoS attacks arising from the feature. In this case DoS protection is provided
by replacing only non-final TXes with final TXes, meaning at most a TX can be replaced once.
As non-final TXes are typically not relayed, concerns about DoS attacks are I think overestimated.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -869,6 +869,7 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
         return false;
 
     // Check for conflicts with in-memory transactions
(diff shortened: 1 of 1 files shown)


rnicoll:
Bitcoin Core have a better patch for this, #bitcoin/bitcoin#5521 , which rejects non-final TXes pre-emptively, so replacement isn't required.

PR #786, opened 2015-01-17
Openssl 1.0.1l dependency update

rnicoll:
Replaces #785 with OpenSSL 1.0.1l version.


langerhans:
Intermediate hashes match :) Thanks!

PR #789, opened 2015-01-28
Block database checks

rnicoll:
Merges in block database checks from Bitcoin Core 0.10


langerhans:
Now that the other is in, this follows

PR #790, opened 2015-01-28
Add -rpcbind option to allow binding RPC port on a specific interface

rnicoll:
Applies changes from bitcoin/bitcoin#3695 into Dogecoin 1.9


langerhans:
Looks good on my Ubuntu 15.04. Bound to localhost and my internal network both fine.  
On a side note, might be worth trying to add the ypthon test to travis.

PR #793, opened 2015-01-29
Fix conflicting declarations via Boost on FreeBSD

### LINKED ISSUE #787: [COMPATIBILITY] For FreeBSD, rename le32enc/dec and be32enc/dec
On FreeBSD, Boost's inclusions of le32enc, le32dec, be32enc and be32dec conflict with Dogecoin's own declarations present in scrypt.cpp and scrypt.h, resulting in a failure to compile.

Suggestion: rename Dogecoin's declarations, perhaps by prefixing as scrypt_le32enc or similar.

stolendata:
Fixes #787


langerhans:
Thanks for the rebase. Looks good to me, as we usually never gonna touch these files again anymore anyway ;)

PR #797, opened 2015-01-29
Set to show title in main window at shutdown.

Sporklin:
Makes Dogecoin show up on shutdown in title thingy ma bobber.


Sporklin:
Rebased

langerhans:
Does what it says :)

PR #801, opened 2015-01-30
Make links in About Dogecoin Core clickable.

Sporklin:
Added clickable thingys. (Something about telling QT to parse some of the HTML)

Devsnack pls?

### DIFF
--- a/src/qt/forms/aboutdialog.ui
+++ b/src/qt/forms/aboutdialog.ui
@@ -113,7 +113,7 @@ This is experimental software.
 
 Distributed under the MIT/X11 software license, see the accompanying file COPYING or &lt;a href=&quot;http://www.opensource.org/licenses/mit-license.php&quot;&gt;http://www.opensource.org/licenses/mit-license.php&lt;/a&gt;.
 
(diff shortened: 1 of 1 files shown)


langerhans:
Missing a space there between the words.
>Thomas Bernard.Paper wallet art provided by Anacoluthia.

Sporklin:
Maybe that works...?

langerhans:
Yup, that works, thanks 👍

Sporklin:
Rebased

langerhans:
It marks them as links, but they don't actually open a browser for me. Tested on Ubuntu 14.04 with FIrefox. If someone else wants to test? If not, might just not pull it.

Sporklin:
It was on the doom sheet lol

rnicoll:
Tested with Ubuntu 14.04 and get the same result. I think this is a bug in QT, however, and personally I'd rather merge this change and keep in sync with Bitcoin Core. I'm also optimistic we'll see this behaviour change in the longer term.

langerhans:
Sounds good to me. Maybe it works in windows. But nontheless, it allows to right click and "copy link location" so I'm fine either way.

PR #802, opened 2015-01-30
Update README.md

leuqarte:
-Change version and copyright year
-rpm-based distribution qt package equivalent to libqtgui4


langerhans:
Thanks, but 1.9-dev is the branch you're looking for ;)

PR #803, opened 2015-01-30
Update README_windows.txt

leuqarte:
-version, copyright year


langerhans:
Same as #802 applies :)

PR #808, opened 2015-01-30
Changes to the Debug window | fix travis

langerhans:
Pulled 3 relating commits into this one PR to have the order right and also fix Travis builds in one. Build with `--disable-wallet` works fine locally, now let's see if Travis wants to complain...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ compiler:
   - gcc
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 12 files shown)


langerhans:
Added a workaround to fix travis for now, until the OpenSSL fix from 1.8.2 is pulled over

langerhans:
split-test-success-kid-meme-300x300

PR #817, opened 2015-01-30
move bignum.h to test/

Sporklin:
tied to #816


langerhans:
Works fine for me after #816. Gonna manually merge this for why ever github is not smart enough :/

PR #820, opened 2015-01-31
Remove template matching params from GetOpName()

Sporklin:
Since they are not real opcodes, being reported as OP_UNKNOWN is less confusing for human-readable decoding.


langerhans:
Restarted travis for the qa-tests...

PR #827, opened 2015-02-01
Add unconfirmed and immature balances to getwalletinfo.

IeuanG:
As shown here bitcoin/bitcoin/pull/5540


IeuanG:
Oh for gods sake, ignore.

PR #833, opened 2015-02-01
estimatefee / estimatepriority RPC methods

rnicoll:
Opening for discussion, DO NOT MERGE.

This is a rework of bitcoin/bitcoin#3959 for Dogecoin. The CFeeRate class changes introduce differences in how transaction fees are calculated, and I've intentionally let the unit tests for transaction fees break here. I'm looking for opinions on whether we should retain transaction fees as-is in Dogecoin (min 1 DOGE, no fine grain fees), or adopt the finer-grained TX fees that this patch introduces to Bitcoin Core.

Personally I favour staying as-is, but wanted to have others chip in. I'll then update the unit tests to match, make code changes as required, and finish testing.

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -1,0 +1,21 @@
+(note: this is a temporary file, to be added-to by anybody, and moved to
+release-notes at release time)
+
+New RPC methods
+===============
+
+Fee/Priority estimation
+-----------------------
+
+estimatefee nblocks : Returns approximate fee-per-1,000-bytes needed for
+a transaction to be confirmed within nblocks. Returns -1 if not enough
+transactions have been observed to compute a good estimate.
+
+estimatepriority nblocks : Returns approximate priority needed for
+a zero-fee transaction to confirm within nblocks. Returns -1 if not
+enough free transactions have been observed to compute a good
+estimate.
+
+Statistics used to estimate fees and priorities are saved in the
+data directory in the 'fee_estimates.dat' file just before
+program shutdown, and are read in at startup.
--- a/qa/rpc-tests/smartfees.py
+++ b/qa/rpc-tests/smartfees.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python
+
+#
+# Test fee estimation code
+#
+
+# Add python-bitcoinrpc to module search path:
+import os
+import sys
+sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), "python-bitcoinrpc"))
+
+import json
+import random
+import shutil
+import subprocess
+import tempfile
+import traceback
+
+from bitcoinrpc.authproxy import AuthServiceProxy, JSONRPCException
+from util import *
+
+
+def run_test(nodes, test_dir):
+    nodes.append(start_node(0, test_dir,
+                            ["-debug=mempool", "-debug=estimatefee"]))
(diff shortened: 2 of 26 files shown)


rnicoll:
bitcoin/bitcoin#4314 is also relevant here, and I'll try to merge it into this PR once we have a decision

rnicoll:
In hindsight, actually, I think lets keep the 1 DOGE min fee for _all_ transactions (remove the free section), but otherwise I quite like the improved granularity on TX fees.

langerhans:
Agree with keeping our current fee structure. It works the way it does. Also it's easy enought to write down into one sentence ;) And no free transactions was introduced now a whole while ago already and seing the value of 1 DOGE it's a non-issue I think.  
(Although sending free transactions is till possible, but psst ;D )

patricklodder:
Can we keep this one open for a little so I can do some investigation if this could help solve our growing "utxo problem"?

langerhans:
@patricklodder Sure, no rush with merging it.

patricklodder:
Ok, so I gave this some thought and basically this change would give us 2 issues:

1. We'll probably have to change dust rules if we want this because this will break up more outputs into  fractional `COIN` value. Do we want to change dust rules? I'd personally be happy not to.
2. This will potentially make utxo even smaller and with that grow in volume. I think we're already in need (somewhat, not desperately) of alternative "input selection" policies, like a fifo policy and a smallest-first policy, on top of the current policy. However, the latter would not be used unless we give people that actively consolidate their utxo an incentive (like a tax-break.) which is a really big subject to discuss and I was planning to put all that in a DIP so we can actually discuss this :)

As for the fee calculation engine proposed here, I really like straight-forward fees as we (kind-of, miners still check coin-age to decide on priority, iirc) have now. `(ceil(bytes.size()/1000) + nSmallOuts)* COIN` is nice, and simple :)

PS: #4305 and #4314 are related to this too.

rnicoll:
Full resync in progress to verify, however this should now behave as expected.

rnicoll:
Think Travis is having hiccups running two tests in parallel, rather than actual problems.

langerhans:
Below from trying this on testnet. Looks kinda odd to me. Maybe it is different on mainnet?  
Feel calculation seems good tho.

image

rnicoll:
The "-1" is not enough information. Syncing on live now to see if it stabilizes in time.

rnicoll:
Estimates are accurate; they reflect expected cost per kb. Given we put a floor in of 1 DOGE, and most transactions are well below 1kb, this gives us estimates in the 3-5 DOGE/kb range for actual cost.

Could come back to this later to change fee calculations to be 1DOGE/kb with a 1 DOGE minimum, but then scaled as per BTC, and have the estimation code special-case 1DOGE fees. For now however, works as expected.

rnicoll:
In fact moments after I write that, I think I have to revise to "Recommend we change calculation to 1 DOGE/kb with a 1 DOGE minimum", to move us closer to BTC calculations, so anything based on BTC maths is readily compatible.

rnicoll:
There, tweaked estimation code to be closer. For some reason values are still high, though.

langerhans:
What we should keep tho is our 1DOGE/ouput<1DOGE extra. Without that, spamming the blockchain would be too cheap...  
Anyway, new estimates look more sane to me, so merging this. Let's try to remmeber to revisit the genral fee structure later after we got the merging done?

PR #840, opened 2015-02-01
 [Qt] allow setting listen via GUI

Sporklin:
- add DEFAULT_LISTEN in net.h and use in the code (shared
  setting between core and GUI)

Important: This makes it obvious, that we need to re-think the
settings/options handling, as GUI settings are processed before
any parameter-interaction (which is mostly important for network
stuff) in AppInit2()!


langerhans:
Works as advertised

PR #843, opened 2015-02-02
qt: PeerTableModel: Fix potential deadlock. #4296

Sporklin:


langerhans:
Tested with DEBUG_LOCKORDER. Looks good.

PR #845, opened 2015-02-02
[Qt] Fix wallet-lock in PaymentServer::fetchPaymentACK(..)

IeuanG:
As merged here bitcoin/bitcoin/pull/5390


langerhans:
Thanks for that. It's way out of order, but checking the upcoming commits, it shouldn't give that much of a conflict. Merging.

PR #852, opened 2015-02-03
bitcoin-cli, rpcclient: prefer EXIT_FAILURE cstdlib constant

Sporklin:
A more complex construction via abs() yields the same end result.

Rebased-From: 34ff109
Rebased-By: Wladimir J. van der Laan <laanwj@gmail.com>


langerhans:
utACK

PR #854, opened 2015-02-03
Add `-stopafterblockimport` option

Sporklin:
Stop after importing blocks. This can be useful for development
and trouble shooting.


langerhans:
Works as advertised

PR #857, opened 2015-02-04
Added 'moon' value in GetInfo()

IeuanG:
Just tested, works fine :+1:


rnicoll:
+UntestedACK

@langerhans ?

langerhans:
utACK, but would like to discuss the term "to the server" there. Are we seing this method only used in daemon and therefore mostly server applications, or do we see Dogecoin Core as client to the Dogecoin network? In latter case we'd have to use a different term there.

IeuanG:
@langerhans Does the term 'to the Dogecoin core' suffice?

langerhans:
Dogecoin **C**ore :)

IeuanG:
Updated @langerhans

patricklodder:
So, under which conditions would we have to patch this to `false`? :)

langerhans:
@patricklodder Probably when safe mode is triggered ;)  
(And in all seriousness, leaving this open until the other changes are merged that touch this file so we don't get too many conflicts in the meantime)

rnicoll:
As we're apparently now actually putting a Dogecoin on the moon, could define this to match whether there is a Dogecoin on the moon. For 1.10.1 though I think, just in case anyone's parsing JSON badly and we break something. Thoughts?

PR #864, opened 2015-02-06
Allow 80 byte OP_RETURN transactions.

Bradley-Plus:
Moved to request pull to 1.9-dev instead of master. Sorry for the confusion.

Bitcoin moved from the 40 byte limit up to 80 byte after seeing no major issues in allowing bigger messages.

I believe this could be valuable in making applications built on top of Dogecoin that may require a teensy bit more than 40 bytes.


Bradley-Plus:
I messed up by putting a tx id in the wrong spot, whoops! Should be fixed now.

PR #868, opened 2015-02-06
Allow 80 byte OP_RETURN transactions.

Bradley-Plus:
I messed up the previous pull requests, but this should be fixed now.

Bitcoin moved from the 40 byte limit up to 80 byte after seeing no major issues in allowing larger storage.

I believe this could be valuable in making applications built on top of Dogecoin that may require a teensy bit more than 40 bytes.

### DIFF
--- a/src/script.h
+++ b/src/script.h
@@ -22,7 +22,7 @@ class CKeyStore;
 class CTransaction;
 
 static const unsigned int MAX_SCRIPT_ELEMENT_SIZE = 520; // bytes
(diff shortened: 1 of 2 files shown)


langerhans:
utACK, since OP_RETURN marks the tx as prunable, so this should not affect our utxo size.

rnicoll:
utACK, but tempted to leave this until late in the 1.9 merge cycle

langerhans:
Already in 1.10 now

PR #870, opened 2015-02-06
Don't allow translation of our example Doge address

Sporklin:
- Changes sample address to address with no held private key.
- Removed unused strings


langerhans:
I tested this on English, de_AT, and Arabic (RTL language). All three look fine. Since it has my commits, waiting for someone else to look over it :)

rnicoll:
Could we get some squashing please? :)

PR #874, opened 2015-02-06
 Push cs_mains down in ProcessBlock

rnicoll:


langerhans:
Testnet sync successful :+1:

PR #876, opened 2015-02-07
[Qt] fix links in about window not opening

Sporklin:
- closes #4402


langerhans:
tested together with #877. Works :)

PR #890, opened 2015-02-09
[Qt] New status bar Unit Display Control and related changes.

Sporklin:
- Removed wasted lines added in original.


langerhans:
Works fine here. Like that feature :)

PR #891, opened 2015-02-09
add OpenSSL RAND_cleanup() on OpenSSL shutdown

Sporklin:
- to securely erase the memory used by the PNRG
- also rework 2 comments in OpenSSL init


langerhans:
utACK

PR #892, opened 2015-02-09
CoinControl related changes after the CFeeRate merge

langerhans:
These kinda complete #833. I explictly set the COIN value as Dust limit here (5d12d1e) as it is one to the user, even if not to the miner.


src/qt/coincontroldialog.cpp
```diff
@@ -448,12 +447,9 @@ void CoinControlDialog::updateLabels(WalletModel *model, QDialog* dialog)
 
         if (amount > 0)
         {
-            if (amount < COIN)
-                fLowOutput = true;
-
             CTxOut txout(amount, (CScript)vector<unsigned char>(24, 0));
             txDummy.vout.push_back(txout);
-            if (txout.IsDust(CTransaction::minRelayTxFee))
```
rnicoll: Why are we moving away from IsDust()?

langerhans: This is just a display thing here. Our IsDust() is hardcoded to false, so the label there would never turn red.  
Our anti dust measure is the 1 doge extra feee. But that is not determined in IsDust().

PR #901, opened 2015-02-10
Replace HexBits with strprintf

Sporklin:
Will not autotest. Needs manual testing.


rnicoll:
Seeing problems with RPC errors with this patch. Re-checking to try isolating cause.

PR #907, opened 2015-02-11
[QT] Replace status bar unit icon with actual images

Sporklin:
- Includes Dogecoin specific changes and related code.
- Added Dogecoin header to bitcoinunits.cpp


langerhans:
Icons look really small on my screen, but that might be due to it beink 4k. Someone with a "normal" screen test please :+1:

rnicoll:
Really small on a normal screen too, sorry @Sporklin

langerhans:
This gets replaced again by bitcoin/bitcoin#5551 . I'll close here and raise a PR for it in a few. Then we have a consistent state there.

PR #931, opened 2015-02-15
[Auto] Don't relay alerts to peers before version negotiation

rnicoll:
Fixes #1436.

Sending alert messages before the negotiation breaks protocol, causing them to be ignored as well as the sending peer to receive a Misbehaving score. This also erroneously prevents the alert from being sent again after version negotiation as it will be already in CNode::setKnown.


rnicoll:
utACK

PR #934, opened 2015-02-15
[Auto] rpc-tests: don't spew non-errors to stdout

rnicoll:
Very minor fix necessary before adding rpc tests to travis.

There's a brief race here, the process might've already exited and cleaned up
after itself. If that's the case, reading from the pidfile will harmlessly
fail. Keep those quiet.


rnicoll:
utACK (arguably tested ACK given it ran fine through Travis)

PR #935, opened 2015-02-15
[Auto] [Wallet] Improve ReorderTransactions(..)

rnicoll:
ReorderTransactions is never called, unless someone would load a very old wallet. I couldnt even trigger ReorderTransactions by experimenting with zapwallettxs. And if it gets triggered somehow, it causes problems, see #3894. The function doesnt even write back txs to disk, if they have nOrderPos == -1, looks like a bug to me.

The worst thing that could happen, is that some txs/entries have nOrderPos = -1. But I dont see any problems with this, even if happen. At least those old transactions would always appear behind new ones.


rnicoll:
utACK - in fact, does this code even apply to Dogecoin? May be worth investigating removal.

PR #936, opened 2015-02-15
[Auto] Add dependencies builder for pull-tester, gitian, easy cross-dev

rnicoll:
This is the first (main) part of a major rewrite of our build and release process: the dependencies builder. See the README in the depends folder for more detailed info. It unifies many of our build procedures so that they may be re-used for similar purposes. The major points are:

- Pull-tester and Gitian build processes have been unified.
- Pull-tester can test all possible configs on all supported OS’s.
- All dependencies are now built with unified recipes. No more hackish per-target shell scripts. This also means that release binaries will share the exact same dependency versions, rather than a random assortment as they are now.
- Gitian builds are now painless and require no intervention since the results are chained together.
- Dependencies are no longer a local issue for the the pull-tester. Everything can be handled in git alone. Anyone can deploy their own pull-tester if desired.
- Devs may elect to use these dependencies rather than installing them via their OS. This is especially helpful on OSX as it avoids the need for homebrew/macports.

Linux x64 and Mac x64 are supported build environments. Building should work from anywhere, it just hasn’t been a priority to actively work outside of those two yet, so the paths haven't been tested.

Build-side tools may either be installed locally or built as native dependencies. For example, the linux/win builds use locally installed gcc/mingw, but the OSX build builds and installs its own clang-based toolchain. Ideally all local build-side dependencies would be removed, but that can come down the road if it proves necessary.

The dependencies builder creates a cached payload for each package that it builds. Each is automatically rebuilt as needed, either when the build recipe or its dependencies are changed. Old payloads are automatically cleaned. The result is a set of files that are automatically glued together into a prefix suitable for building Bitcoin. This process is entirely automated and deterministic, suitable for running before each  Bitcoin build. If nothing has changed, it completes in less than a second.

A POC pull-tester has been setup to use them: https://travis-ci.org/theuni/bitcoin/builds . It caches these packages and re-uses them as necessary (note that caching is not currently happening due to some Travis account restrictions, I’m actively working with the Travis guys on this. Once this is in place builds will only take a few min). The end-result is that Bitcoin is always built with a known set of dependencies, which are entirely controlled by us. No files from the build-machine are ever used. This means that the pull-tester can be OS/distro/architecture agnostic, so we don’t end up in a situation (as we are currently) where the pull-tester is a static environment that can’t be touched.

Gitian has been made to use these dependencies as well. Now, Gitian releases are built with the exact same procedure as the pull-tester, making dogfooding much easier. As mentioned for the pull-tester above, this lifts the OS restriction on Gitian, so we’re free to move to more recent distros/releases. Builds have been verified as deterministic locally, but I haven’t yet compared with non-local builds. Also worth mentioning, because we build all dependencies ourselves in a sandbox, a single base image can be used for all builds.

The new Gitian descriptors have not been included here, they will come in as a separate PR. However, the changes to enable our build-cache have been accepted upstream: https://github.com/devrandom/gitian-builder/pull/62

With this new version, the sources are fetched by the dependencies builder, and cached for re-use. Existing files in the inputs folder will not be used, so the first build will take a good big longer.

TODO:
Enable the travis builds. I'm currently working with Travis on this, I hope to have it ready in the next week.


langerhans:
Working for me with some changes I'll PR in a few. Merging this first.

PR #941, opened 2015-02-15
[Auto] When reindexing check for file before trying to open

rnicoll:
Fix for #4693 -- check for each file before trying to open, so an error is only logged if the file exists but is unreadable.

I thought about refactoring the code so that the way you map from a file number to a path is done in one function only, but the assumption that block files look like "blkNNNNN.dat" is in several places, including comments and debug messages, so wasn't sure such a refactor was warranted compared with this simple fix.


rnicoll:
utACK

PR #942, opened 2015-02-15
[Auto] Also switch the (unused) verification code to low-s instead of even-s.

rnicoll:
a81cd968 introduced a malleability breaker for signatures (using an even value for S). In e0e14e43 this was changed to the lower of two potential values, rather than the even one. Only the signing code was changed though, the (for now unused) verification code wasn't adapted.

Note to @gavinandresen: this brings the total number of test cases to 111.


rnicoll:
Turns out it will actually apply blank commits, sorry

PR #962, opened 2015-02-15
[Auto] Do not flush the cache after every block outside of IBD

rnicoll:
It seems the reason why jenkins has been failing for several months is that a 200-pathological-block reorg takes over 20 minutes, at which time a timeout occurs.

The reason is that since a refactor, we're now doing reorganizations in steps (and since more recently, even releasing cs_main in between if possible), which means that outside of IBD, we're dumping the entire coins database (over a million txids in the large-reorg test!) to disk, and removing it from memory after every block.

This change, in combination with a sufficiently large -dbcache may be enough to bring that time down.


langerhans:
Depends on at least bitcoin/bitcoin@dd638dd712861a2fc419fe9b2b064ae303b074e6

PR #968, opened 2015-02-15
[Auto] Test CHECKMULTISIG with m == 0 and/or n == 0

rnicoll:
Turns out that's valid, and we don't test it.


rnicoll:
Sorry, another empty one...

PR #975, opened 2015-02-15
miner: indentation fixes, remove for (;;)

langerhans:
- change a for (;;) into while (true), as we nowhere else use the first
- init nNonceFound to 0
- fix indentation in BitcoinMiner try/catch block


langerhans:
Tessted by mining 10 blocks on regtest

PR #991, opened 2015-02-19
[Auto] Bitcoin PR batch Thu Feb 19 09:34:02 2015

rnicoll:
Contains:

* Add /opt/local/include/db48 only if it exists.


rnicoll:
Should this be 5.1 instead?

langerhans:
Isn't that taken care of by `dogecoin_find_bdb51.m4` ? But yeah, that should be 51 instead. For consistency.

rnicoll:

PR #994, opened 2015-02-20
[Auto] Bitcoin PR batch Fri Feb 20 06:55:50 2015

rnicoll:
Contains:

* RPC client: Simplify command line string-to-JSON-value conversion code
* Show nodeid instead of addresses (for anonymity) unless otherwise requested
* Coding style update with clang-format
* Use complete path to include bitcoin-config.h.


rnicoll:
Think this needs further work on the RPC client changes, as I don't think it takes into account getauxblock

rnicoll:
Both parameters to getauxblock are strings (unlike getblocktemplate), so that's not an issue.

PR #996, opened 2015-02-20
Fix some python RPC tests

langerhans:
This includes bitcoin/bitcoin@49d86c7477ba1dbbbd08f452d0e7d710b22bad92. `smartfee.py` is still broken though and I can't really get what it's supposed to test anyway. I also hate python :)


langerhans:
Travis cancelled for obvious reasons.

PR #999, opened 2015-02-21
Add unit tests for calculating next difficulty

rnicoll:
Add new unit tests which test difficulty calculation pre and post Digishield works as expected. Requires
some refactoring to isolate the maths from the originating data.


rnicoll:
Tweaked tests to ensure they cover as much of the code as possible, and to identify cases each tests.

langerhans:
Verified the test vectors against my local blockchain. Thanks!

PR #1005, opened 2015-03-08
[Auto] Bitcoin PR batch Sat Mar  7 19:13:16 2015

rnicoll:
Contains:

* replace int with size_t in stream methods
* Clean-up SyncWithWallets/SyncTransaction
* Use explicit fflush() instead of setvbuf()
* Move g_signals.SetBestChain(..) below SyncWithWallets


rnicoll:

PR #1016, opened 2015-03-08
init.cpp: log fee estimates filename on error

Sporklin:
- small changes to Shutdown(), buffer __func__, which is now used in
  all LogPrintf() calls and format for better readability
- order using namespace alpabetically


langerhans:
utACK

PR #1017, opened 2015-03-09
Use unordered_map for CCoinsViewCache with salted hash

Sporklin:


langerhans:
Rebased locally, looks good.

PR #1019, opened 2015-03-09
qt: Move SplashFinished to after ClientModel/WalletModel creation

Sporklin:
With a large wallet there was a noticable gap between hiding of the
splash and showing the main window.


langerhans:
utACK

PR #1020, opened 2015-03-09
Add warning comment to getinfo

Sporklin:
Warn that people should not add new information, or change current
information returned by getinfo.


whitj00:
ut (but i will pay @patricklodder 500k Ðoge if it does break anything) ACK

:octopus:

PR #1026, opened 2015-03-12
add GetRandBytes() as wrapper for RAND_bytes() 

Sporklin:
- add a small wrapper in util around RAND_bytes() and replace with GetRandBytes() in the code to log errors from calling RAND_bytes()
- remove OpenSSL header rand.h where no longer needed
- move rand functions from util to new random.h/.cpp


langerhans:
Works for me

PR #1033, opened 2015-03-13
Travis test fixes

Sporklin:


langerhans:
utACK, we don't use the comparison tool anyway (at the moment at least?)

PR #1035, opened 2015-03-14
Fix the build for windows

Sporklin:
Problem introduced in caf6150. Thanks to @drak for noticing. Fixes #4473.


langerhans:
utACK

PR #1036, opened 2015-03-14
Add helptexts for -whitelist and rpc prioritisetransaction and switch -

Sporklin:
to bitcoin instead of satoshis.

- Dogecoin specific disabled changes accounted for. @rnicoll turned off in #969


langerhans:
Putting  https://github.com/bitcoin/bitcoin/pull/5398 here for reference.

PR #1041, opened 2015-03-15
[Qt] format ping times in peers tab as ms

Sporklin:
- also align ping times to the right


langerhans:
restarted travis

langerhans:
utACK

PR #1043, opened 2015-03-15
net: add SetSocketNonBlocking() as OS independent wrapper

Sporklin:


langerhans:
Synced some blocks, looks good

PR #1044, opened 2015-03-15
Clean up RPCs that are disabled in safe-mode.

Sporklin:
This removes some inconsistencies in what worked and didn't work in
 safemode. Now only RPCs involved in getting balances or sending
 funds are disabled.

Previously you could mine but not submit blocks— but we may need more
 blocks to resolve a fork that triggered safe mode in the first place,
 and the non-submission was not reliable since some miners submit
 blocks via multiple means. There were also a number of random commands
 disabled that had nothing to do with the blockchain like verifymessage.

Thanks to earlz for pointing out that there were some moderately cheap
 ways to maliciously trigger safe mode, which brought attention to
 the fact that safemode wasn't used in a very intelligent way.


langerhans:
utACK

PR #1052, opened 2015-03-16
 prioritisetransaction RPC: Restore compatibility with existing implemen...

Sporklin:
...tations by using satoshis for fee offset rather than BTC

### DIFF
--- a/src/rpcmining.cpp
+++ b/src/rpcmining.cpp
@@ -285,6 +285,7 @@ Value getmininginfo(const Array& params, bool fHelp)
 }
 
 
+// NOTE: Unlike wallet RPC (which use BTC values), mining RPCs follow GBT (BIP 22) in using satoshi amounts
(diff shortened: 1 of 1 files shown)


src/rpcmining.cpp
```diff
@@ -296,14 +297,14 @@ Value prioritisetransaction(const Array& params, bool fHelp)
             "2. priority delta (numeric, required) The priority to add or subtract.\n"
             "                  The transaction selection algorithm considers the tx as it would have a higher priority.\n"
             "                  (priority of a transaction is calculated: coinage * value_in_satoshis / txsize) \n"
-            "3. fee delta      (numeric, required) The absolute fee value to add or subtract in bitcoin.\n"
+            "3. fee delta      (numeric, required) The fee value (in satoshis) to add (or subtract, if negative).\n"
```
rnicoll: Can this be "in koinu" instead of "in satoshis" please?

PR #1064, opened 2015-03-17
build: check for sys/prctl.h in the proper way

Sporklin:
Use AC_CHECK_HEADERS to check for the header, and include it only
if detected and the subsequent HAVE_SYS_PRCTL_H is set.


langerhans:
utACK, since travis is happy

PR #1066, opened 2015-03-18
 Create new signal for notification of new blocks. Use w/ -blocknotify

Sporklin:


langerhans:
ACK together with #943

PR #1068, opened 2015-03-18
Clean-up SyncWithWallets/SyncTransaction

Sporklin:


rnicoll:
I note with slight alarm that the original PR ( bitcoin/bitcoin#4716 ) appears to only have untested ACKs! Anyway, tested, seems okay.

PR #1075, opened 2015-03-18
qt: fix unicode character display on osx when building with 10.7 sdk

Sporklin:


langerhans:
I don't have a 10.7 SDK, and we probably will upgrade later on anyway if I'm not mistaken, so utACK.

PR #1078, opened 2015-03-18
dogecoin-tx: fix build warnings

Sporklin:
Cleans up a bunch of:
warning: missing braces around initializer for ‘const<anonymous struct>’


langerhans:
utACK

PR #1079, opened 2015-03-19
fix compilation error in core_io.h

Sporklin:
- error: 'vector' in namespace 'std' does not name a type
- add <vector> include in core_io.h
- remove <vector> includes from core_read.cpp and core_write.cpp


langerhans:
utACK, travis is happy

PR #1082, opened 2015-03-19
univalue: remove unused methods getArray, getObject

Sporklin:


langerhans:
utACK, travis is happy

PR #1084, opened 2015-03-19
Add "it works" test for dogecoin-tx 

Sporklin:


langerhans:
`PASS: test/bitcoin-util-test.py`

Sounds good to me :)

PR #1089, opened 2015-03-19
build: fix automake warnings about the use of INCLUDES

Sporklin:
While we're at it, reduce the use of LIBS as well. This makes dependencies
explicit.

Fixes building with (the not-yet-merged) libsecp256k1 as well.


langerhans:
Tested that before the PR went in, travis happy too.

PR #1090, opened 2015-03-21
Make Get/Have in CCoinsView "const"

Sporklin:
This patch makes certain functions (Read/Exists in CLevelDBWrapper and diverse Get/Have ones in CCoinsView and subclasses) const. IMHO, this is the "right thing" for them, since they represent in fact "read-only access". This can also be used to make the CCoinsView reference in some higher-level functions (e. g., CheckInputs) const.

Is there a reason why this can't be done? It should make the code cleaner. Note that the three commits encapsulate "different levels" in the change here, and only the first two or only the first can be used, too, in case there is a good reason why the full patch isn't appropriate.

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -52,30 +52,30 @@ bool CCoins::Spend(int nPos) {
 }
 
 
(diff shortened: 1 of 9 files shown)


langerhans:
Seems to work fine here.

PR #1094, opened 2015-03-22
[Qt] move SubstituteFonts() above ToolTipToRichTextFilter

Sporklin:
- doesn't belong to the ToolTipToRichTextFilter class so move it up


langerhans:
utACK

PR #1096, opened 2015-03-23
Added formal english

MachoSmurf:
Copied English language file and removed dogespeak.

Someone with more knowledge about the QT client should probably make an option to enable "Dogespeak". In my opinion normal English with a normal font would be best to have as a default. There was some discussion within the community as well. Rnicoll suggested this translation to be made into a pull request so it could be discussed here as well.

### DIFF
--- a/src/qt/locale/bitcoin_en_formal.ts
+++ b/src/qt/locale/bitcoin_en_formal.ts
@@ -1,0 +1,4297 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.0" language="en">
+<context>
+    <name>AboutDialog</name>
+    <message>
+        <location filename="../forms/aboutdialog.ui" line="+14"/>
+        <source>About Dogecoin Core</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location line="+39"/>
+        <source>&lt;b&gt;Dogecoin Core&lt;/b&gt; version</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location line="+57"/>
+        <source>
+This is experimental software.
+
+Distributed under the MIT/X11 software license, see the accompanying file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+This product includes software developed by the OpenSSL Project for use in the OpenSSL Toolkit (http://www.openssl.org/) and cryptographic software written by Eric Young (eay@cryptsoft.com) and UPnP software written by Thomas Bernard.</source>
+        <translation>
+This is experimental software.
+
+Distributed under the MIT/X11 software license, see the accompanying file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+This product includes software developed by the OpenSSL Project for use in the OpenSSL Toolkit (http://www.openssl.org/) and cryptographic software written by Eric Young (eay@cryptsoft.com) and UPnP software written by Thomas Bernard.</translation>
+    </message>
+    <message>
+        <location filename="../utilitydialog.cpp" line="+29"/>
+        <source>Copyright</source>
+        <translation>Copyright</translation>
+    </message>
+    <message>
+        <location line="+0"/>
+        <source>The Dogecoin Core developers</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location line="+12"/>
+        <location line="+2"/>
+        <source>(%1-bit)</source>
+        <translation type="unfinished"></translation>
+    </message>
+</context>
+<context>
+    <name>AddressBookPage</name>
+    <message>
+        <location filename="../forms/addressbookpage.ui" line="+30"/>
+        <source>Double-click to edit address or label</source>
+        <translation>Double-click to edit address or label</translation>
+    </message>
+    <message>
+        <location line="+27"/>
+        <source>Create a new address</source>
+        <translation>Create a new address</translation>
+    </message>
+    <message>
+        <location line="+3"/>
+        <source>&amp;New</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location line="+11"/>
+        <source>Copy the currently selected address to the system clipboard</source>
+        <translation>Copy the currently selected address to the system clipboard</translation>
+    </message>
+    <message>
+        <location line="+3"/>
+        <source>&amp;Copy</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location line="+58"/>
+        <source>C&amp;lose</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location filename="../addressbookpage.cpp" line="+74"/>
+        <source>&amp;Copy Address</source>
+        <translation>&amp;Copy Address</translation>
+    </message>
+    <message>
+        <location filename="../forms/addressbookpage.ui" line="-47"/>
+        <source>Delete the currently selected address from the list</source>
+        <translation>Delete the currently selected address from the list</translation>
+    </message>
+    <message>
+        <location line="+27"/>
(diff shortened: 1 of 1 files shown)


patricklodder:
Thank you for taking initiative on this.

I think I agree with you re: proper english should be the default, however that would be done (properly) by changing the sources, not the translation file. We can do that before my next planned bitcoin_en.ts sync, which I planned to do around the beginning of next month. 

@langerhans & @rnicoll: thoughts?

langerhans:
To be honest, I'd rather keep the default as it is, but provide a setting for at least the language and the font (Could be extended to some of the icons too). Dogecoin is Dogecoin and I feel like it should stay true to its roots. That being said I did have that on my to do list, imagining it as setting on the first start disalog. It would allow setting the language and the font since those are fairly easy to change. But that would happen after all the merging work is done. 

To give some more scale to it: Even MultiDoge had Comic Sans as default font for a while. But since that Java font system is a bit bugged I had to revert that. It's still an option there. No Dogespeak there pretty much amounts to my own laziness back in the days. 

Hating on Comic Sans is just "in" cause it has been done forever. Blaming Dogespeak and the font less acceptance though is wrong in my view, as the first thing people see is certainly **not** the reference client, but the website and the communities. I think the website does quite a good job so far. But Would you want to force all the communities to remove Dogespeak to please your target audience for this change?

MachoSmurf:
I think that the difference with the website and the community is that there is a relatively low amount of Dogespeak, yet the wallets main functions are all in dogespeak, unless you switch to another language because those are (as far as I know) in normal language. That's also part of the reason I suggested this for English.

I can understand that someone who is completely new to the concept of cryptocurrencies can be frightened by this weird "English dialect". Once someone has got the basics down and gets some knowledge about the doge meme it's no problem to switch. So I think the solution you suggest is a perfect compromise.

rnicoll:
I thought there was a move towards Comic Neue anyway, as a bit clearer?

I'd like to have the playful font & wording as defaults, but personally I think we're suffering for it. I think we do need to move to at least a clearer font, and vanilla language as defaults, and add a "Dogecoin Classic" option to make it playful again.

patricklodder:
Re: font. To me the current font is not an issue except i think there was a reason for Comic Neue because of licensing/portability? Either way for me the font can stay as that is non-obstructive as long as it's legal and deployable multi-platform.

Re: wording. Regardless of what is deemed wisdom regarding default language, I strongly feel that "Dogespeak" should not be the source language (as it is sitting around QT source files) but plain English with a `.ts` file that contains the playful wording. This ts could be the default, or it could not, that really doesn't matter. We're currently asking for translations of "Dogespeak" sentences and those do not match the delivered translations; not for the ones we retrieve from upstream nor the ones that have been maintained manually. It does not encourage someone to translate "very sending addresses" because at best the translation would be even more confusing.

Re: which wording to use by default. I hang a little towards @rnicoll's feeling that we're at that point where our coin has matured past being a joke and we might (I don't have much data on the subject so I can't be sure) be suffering in terms of adoption because of the odd wording.

However, I also see that 1.8.x releases have had much less downloads than 1.6.x (250k- vs 1M+), so the exposure of Dogecoin Core seems to be shrinking over the past 8 or perhaps 10 months, while during the same period, web wallets have been picking up and those are attracting people from outside the Reddit crowd mostly. So there might be an obstruction to use Dogecoin Core, question is: is it the "DogeSpeak".

I personally don't expect the number 1 obstruction to be the language used as default but rather the parsing of the entire chain and all the little inconveniences that come with that. Even if it would come down to UI, improvements in other areas (as in general user experience) might have greater effects than changing the language, so if we think we're hurting, that's all cool, but this is probably not the main cause, in my opinion.

TLDR: 
- making `bitcoin_en.ts` vanilla English: yay. 
- making a separate `.ts` for dogespeak: yay. 
- replacing comic sans with verdana: nay. 
- replacing comic sans with something that looks almost the same but is more open/free/portable: yay.
- making dogespeak or vanilla english the default: really don't care.

ummjackson:
Why so serious?

Eh, my vote is that we keep Comic Sans + dogespeak. Never lose sight of the fact that Dogecoin is about fun/absurdity.

michilumin:
I don't think anyone's "losing sight" of that fact, Jackson, just wanting to clean up a little. But I have two thoughts on that;  One, that it's not really necessary to limit Dogecoin to never be able to serve purposes beyond the absurd, (in fact, it kind of carries its "happy spirit" into the serious transactions it's involved in, rather than Dogecoin being brought down to the level of other cryptos -- which leave a bad taste in my mouth, and i'm sure many others mouths as well.)   it can actually bring its levity into situations that don't have it.  I think that Dogecoin elevates the banal up to the happy, versus the other way around.  The reason it does that, at this point, is much bigger than comic sans.

My other thought is that by giving it a refresh, we get a few more inroads into Dogecoin being taken just seriously enough to be used.   "Fun and absurd" (I'd like to say fun and light-hearted) doesn't necessarily mean kid scribbles and comic sans.  There's nothing saying that changing one will destroy the other. 

Jackson, I know your drumbeat is, "it's a stupid joke, it's a stupid joke" with a quiet unspoken commentary of "and anyone who tries to let it evolve beyond that is an idiot".  I get that. You created it, that's fine, but it'd be real hard to say that to some degree Dogecoin hasn't taken on a life of its own well past the prime of the meme. And yes, has outlived your intentions for it even! And there are people who want it to, for one, stay alive and somewhat relevant, and two, continue to gain acceptance. I know I'm probably going to be beat over the head by merely suggesting that.

So I'm absolutely going to throw my hat in the ring with @patricklodder and @rnicoll , saying that while the "Ha-ha it's just a joke guys!" time was fun, a lot of work has been put in both community and code-wise to take it a bit beyond that, and it's that community, spirit, and force (not the lols/joke/meme alone) that has allowed it to persist where other altcoins have failed.   

Dogecoin has survived a lot and it should be allowed to grow up just a little. Not a lobotomy. Not a personality change. Just a bit of evolution. 

Mostly plain English with a phrase here or there of dogespeak, I think would be great. A slightly more refined font that still says playful. I'd love to see the look of the client updated too, making the application itself look more modern, and I'd be fully willing to work on it. (I was in the past, but was warned that it wouldn't be well received as it's supposed to stay kind of 'dumpy and silly'. I'm glad this discussion is happening.) 

I know that's just a bit of a "me too", and a "who the hell are you?" said some things counter to the great @ummjackson - but, welp, there it is.

Sporklin:
It is simple. I do not really care for the why and the excuses for it honestly. For whatever maturity and levity it could be argued that the Dogespeak and the font add or remove, it is highly irrelevant given we are still getting users, we have users..Shockingly enough we got them, are continuing to get them with things being what they are.  Dogecoin is what it is and has done what it has due in part to the ideals around it. Dogecoin is not validated and based on one, two, even three minor cosmetic things.

This said there is an issue with the Dogespeak. It is the source from which everything is translated, this has led to some overly confusing, misleading and direct translations of some words that are causing issues in the communities that do not speak English as a primary or solid secondary language. Where they lack the context of intention with some of the words, given there is no real way to translate or convey the meaning, so there is gibberish coming out.

Have formal English be an option have it also be the source for where translation is pulled from. Dogespeak stays at default, with the font up until which point it gets too restrictive to have comic sans be the font in which cause comic neue does look nice.

It is win win.

Dev TLDR: I do PRs now, you can read what I said.

rnicoll:
Still slightly chewed from asking the community about Comic Neue, I'm inclined to say thanks for the work, but we can't really merge this. @langerhans , @patricklodder ?

patricklodder:
I still don't mind having this as an option, in fact, i'd still like the strings in the source code to be formal english, with only dogespeak as the default translation.

rnicoll:
Bumping to 1.10.1. as we need 1.10 to be non-controversial for adoption, but will hammer this into shape for that release then.

PR #1099, opened 2015-03-23
build: work around ccache/autotools warning-spamming bug

Sporklin:
When using clang and ccache, builds spew lots of:
Clang: warning: argument unused during compilation

Upstream bug: https://bugzilla.samba.org/show_bug.cgi?id=8118

This is harmless, bug annoying. If ccache is being used and the
-Qunused-arguments flag is supported (clang), use it.


langerhans:
utACK, merging as prereq for travis work

PR #1101, opened 2015-03-23
qt/splashscreen: #include version.h

Sporklin:
Needed to build breakage reported by Arnavion on IRC:
qt/splashscreen.cpp: In constructor 'SplashScreen::SplashScreen(const QPixmap&, Qt::WindowFlags, bool)':
qt/splashscreen.cpp:33:98: error: 'FormatFullVersion' was not declared in this scope


langerhans:
utACK

PR #1105, opened 2015-03-23
Depends fixes round1

Sporklin:


langerhans:
Can't really test that, but looks fine code wise.

PR #1107, opened 2015-03-24
Univalue: Do not build JSON escape list at runtime

Sporklin:
No need to waste startup time building something that can be done
at compile time.

This also resolves a clang++ warning originally reported in #4714,
univalue/univalue_write.cpp:33:12: warning: array subscript is of type 'char
escapes['"'] = "\\"";
^~~~

etc.


langerhans:
utACK

PR #1112, opened 2015-03-24
Closely track mempool byte total. Add "getmempoolinfo" RPC.

Sporklin:
Goal:  Gain live insight into the mempool.  Groundwork for future work
that caps mempool size.

- Fee changes related, expect conflicts when merging.

### DIFF
--- a/src/rpcblockchain.cpp
+++ b/src/rpcblockchain.cpp
@@ -567,3 +567,27 @@ Value getchaintips(const Array& params, bool fHelp)
 
     return res;
 }
+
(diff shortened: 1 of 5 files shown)


langerhans:
Works as advertised locally. Waiting on at least travis (and maybe we get the fee rate change in too?)

langerhans:
Needs rebase, also should now be able to include the full merge with the fee changes.

langerhans:
Now in #1166 due to misunderstanding. Closing here

PR #1127, opened 2015-04-25
Smart fees for wallet CreateTransaction

rnicoll:
Dogecoin equivalent of bitcoin/bitcoin#4250

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -1,6 +1,26 @@
 (note: this is a temporary file, to be added-to by anybody, and moved to
 release-notes at release time)
 
+Transaction fee changes
+=======================
+
+This release automatically estimates how high a transaction fee (or how
+high a priority) transactions require to be confirmed quickly. The default
+settings will create transactions that confirm quickly; see the new
+'txconfirmtarget' setting to control the tradeoff between fees and
+confirmation times.
+
+Prior releases used hard-coded fees (and priorities), and would
+sometimes create transactions that took a very long time to confirm.
+
+
+New Command Line Options
+========================
+
+-txconfirmtarget=n : create transactions that have enough fees (or priority)
+so they are likely to confirm within n blocks (default: 1). This setting
+is over-ridden by the -paytxfee option.
+
 New RPC methods
 ===============
 
--- a/src/wallet.h
+++ b/src/wallet.h
@@ -267,6 +267,7 @@ public:
     std::string SendMoney(CScript scriptPubKey, int64_t nValue, CWalletTx& wtxNew);
     std::string SendMoneyToDestination(const CTxDestination &address, int64_t nValue, CWalletTx& wtxNew);
 
+    static CFeeRate minTxFee;
     static int64_t GetMinimumFee(const std::vector<CTxOut> &vout, unsigned int nTxBytes, unsigned int nConfirmTarget, const CTxMemPool& pool);
 
     bool NewKeyPool();
--- a/src/wallet.h
+++ b/src/wallet.h
@@ -25,12 +25,14 @@
 
 // Settings
 extern CFeeRate payTxFee;
+extern unsigned int nTxConfirmTarget;
 extern bool bSpendZeroConfChange;
 
 // -paytxfee default
 static const int64_t DEFAULT_TRANSACTION_FEE = 0;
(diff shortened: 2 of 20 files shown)


rnicoll:
For discussion and review, do not merge yet. Will test further this afternoon.

src/wallet.h
```diff
@@ -25,12 +25,14 @@
...
 // Settings
 extern CFeeRate payTxFee;
+extern unsigned int nTxConfirmTarget;
 extern bool bSpendZeroConfChange;
 
 // -paytxfee default
 static const int64_t DEFAULT_TRANSACTION_FEE = 0;
 // -paytxfee will warn if called with a higher fee than this amount (in satoshis) per KB
 static const int nHighTransactionFeeWarning = 25 * COIN;
+static const unsigned int MAX_FREE_TRANSACTION_CREATE_SIZE = 1000;
```
patricklodder: propose: MAX_FREE_TRANSACTION_CREATE_SIZE = 0;

rnicoll:
That should be it ready to merge now unless there's any other feedback.

doc/release-notes.md
```diff
@@ -1,6 +1,26 @@
...
+high a priority) transactions require to be confirmed quickly. The default
+settings will create transactions that confirm quickly; see the new
+'txconfirmtarget' setting to control the tradeoff between fees and
+confirmation times.
+
+Prior releases used hard-coded fees (and priorities), and would
+sometimes create transactions that took a very long time to confirm.
+
+
+New Command Line Options
```
langerhans: Reminder: Rework the release notes at release time, as this is not actually in with this PR

langerhans:
I'll give this a try soon, but looks good so far.

langerhans:
Seems like it doesn't calculate the dust fee correctly. Sent this tc on mainnet just now: 3d6ec3ae2aca3ae0a6c65074fd8ee888cd7ed262f2cbaa25d33861989324a14e  

It should have been 2 DOGE fee because of the dust output, no?

rnicoll:
Think so, but can't see how it would miss it, and more I can't see how it was relayed if it didn't meet the dust requirements :-/

rnicoll:
Not going to be able to find an answer on this tonight. I've tracked up and down wallet.cpp, can't see any problems, if you have any ideas let me know

rnicoll:
Added specific unit test for the wallet fee calculation, still no sign of a problem :(

rnicoll:
Think I may have found the problem - tests only cover calculating the fee, while below-fee transactions could still get through the wallet. Pushing this to get it off an EC2 node, but haven't finished testing, will give a thumbsup when done.

rnicoll:
Looks good to me, now: https://chain.so/tx/DOGETEST/d9195e65bbe92a4bb63eaf267ce8d378a7d6c626a340fab24d27d24425bfb4f6

langerhans:
Odd, this is what I see. Clean build with this PR merged in.

image

Sample transaction: https://chain.so/tx/DOGE/cb505cdf8f20041f86338041d8ef36068d5c5abe82ceab9795a6f026352aa541  

I'll have to sync in Linux to actually try and debug it.

rnicoll:
You've got the commits from this morning included, right?

I'll try re-testing in the morning, always possible I've got versions muddled...

langerhans:
I did, I re-checkoud out your whole PR. But I'll check that too :)

rnicoll:
Found a few more places that references free transactions, ripped them out to be sure, but seems okay here *confused*

untitled

langerhans:
@rnicoll It seems to behave differently on Testnet and Mainnet. On Testnet I can reproduce your results, on Mainnet I get 1 DOGE for a 0.5 DOGE send.

rnicoll:
Thought I'd disabled the fee estimation from existing transactions, but might be it's seeing most transactions of the same size pay 1 DOGE, and using that...

rnicoll:
Nope, txconfirmtarget is definitely disabled. I'm open to suggestions...

langerhans:
Found the issue: I had a 1 DOGE fee set in the settings. That means this settings fully overrides any fee estimation. I'm not so sure if that is intended to be the way, but we can check on that later. I'll merge this now, so that holdup is finally out of the way.

PR #1136, opened 2015-06-19
advertise block height addition to -blocknotify

stolendata:


langerhans:
Closing since #1238 replaced it. Thanks tho!

PR #1138, opened 2015-06-19
univalue: make spaceStr thread-safe

Sporklin:
Simply add spaces to the existing string instead of using a
temporary.

Fixes bitcoin/bitcoin#4756.


langerhans:
utACK, we have literally no own changes there.

PR #1139, opened 2015-06-19
add missing header end comments

Sporklin:
- ensures a consistent usage in header files
- also add a blank line after the copyright header where missing
- also remove orphan new-lines at the end of some files


langerhans:
utACK

PR #1140, opened 2015-06-19
build: osx: Fix incomplete framework packaging for codesigning

Sporklin:
Starting with 10.9, Framework versions must be signed individually, rather
than as a single bundle version, in order to be properly codesigned. This
change ensures that the proper plist files and symlinks are present prior to
packaging.


langerhans:
Merging untested, as it seems to be universal enough for us. We don't have changes on that script anyway.

PR #1141, opened 2015-06-19
Fix build with DEBUG_LOCKORDER

Sporklin:
Fixes bitcoin/bitcoin#4771


langerhans:
Compiled with `-DDEBUG_LOCKORDER` and works.

PR #1143, opened 2015-06-20
Move g_signals.SetBestChain(..) below SyncWithWallets

Sporklin:


langerhans:
utACK for the move. Although the magic numbers there are odd for us. We've been running fine with them for so long though, so I don't see a need to touch them.

PR #1145, opened 2015-06-20
Bugfix: bitcoin-tx: scriptPubKey may be null, so accept outscript=<n>:

Sporklin:


langerhans:
ACK, cause travis tested it already.

PR #1166, opened 2015-06-21
Closely track mempool byte total. Add "getmempoolinfo" RPC.

Sporklin:
Goal:  Gain live insight into the mempool.  Groundwork for future work
that caps mempool size.


langerhans:
This was tested in #1112 already, so merging.

PR #1167, opened 2015-06-21
Fix feeestimate

Sporklin:
Fixed a bug with index bounds checking.
Fix minor bug which only affected log messages.
Process fee estimate file into temporary vector first to let sanity checking complete.


langerhans:
Tested locally. On Mainnet fee estimates look about the same as before. Code looks okay too to me, so we can merge this.

PR #1168, opened 2015-06-21
Add detailed network info to getnetworkinfo RPC

Sporklin:
This commit adds per-network information to the
getnetworkinfo RPC call:

- Is the network limited?
- Is the network reachable
- Which proxy is used for this network, if any

Inspired by bitcoin/bitcoin #2575.


langerhans:
Does what it say, tested on regtest.

PR #1173, opened 2015-06-22
build: fix osx+homebrew+qt5 builds

Sporklin:
Qt5 is keg-only, so configure won't find it without some help. Use brew to find out its prefix.

Also, qt5 added the host_bins variable to pkg-config, use it.

Language change to homebrew included as well.

*Dogecoin specific db change

(No yelling at me for changed wording on the thingy >_> Blame @langerhans he said to change it)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -236,7 +236,7 @@ case $host in
 
        AC_CHECK_PROG([BREW],brew, brew)
        if test x$BREW = xbrew; then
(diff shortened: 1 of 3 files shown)


DomT4:
> Qt5 is bottled

`keg-only` :wink:. Bottles are our CI produced binaries, `keg_only` is the process of locking something within its prefix and not symlinking it into the common prefix, i.e. `/usr/local` usually.

PR #1175, opened 2015-06-23
[QT] small Qt-only include cleanup

Sporklin:
* Excluded changes in transactiontablemodel.cpp due to the choice made not to implement SI-style. See #822

### DIFF
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -15,11 +15,11 @@
 #include "optionsmodel.h"
 
 #include "main.h" // for MAX_SCRIPTCHECK_THREADS
+#include "netbase.h"
+#include "txdb.h" // for -dbcache defaults
 #ifdef ENABLE_WALLET
 #include "wallet.h" // for CWallet::minTxFee
 #endif
-#include "netbase.h"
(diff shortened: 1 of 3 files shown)


prestotron:
It seems you just relocated things. It doesn't make much of a difference whether you "cleaned" it up or not.

I guess it wouldn't hurt to merge this, but just saying it doesn't make a difference :P

Thanks for cleaning up though, organization is always welcome :)

Sporklin:
It is a cherry picked commit from the Bitcoin upstream. Important because we follow theirs, so their changes even the small ones are important to have in order to make merges easier as we go along.

prestotron:
@Sporklin I guess that is true. I just don't see why it makes a difference. 

We'll see how it goes.

langerhans:
@prestotron We are currently in the process of merging all changes from upstream bitcoin. So small commits like this pop up eventually and quite often.  
It make a difference, as stying up to date with the codebase from upstream leaves us with less merge conflicts later down the line. While it might seem pointlesss to you, it very much has a point ;)

prestotron:
@langerhans Ah okay, thank you for explaining this. 

I didn't understand beforehand. At least I know have some ideas now for new pulls :)

PR #1176, opened 2015-06-23
RAM suggestions and Swap

prestotron:
Some people have issues compiling or running because of RAM. I added a "suggested ram" thingy to 4GB, and directed users to a tutorial on how to add swap if it was not working with below 4GB of RAM.

Nothing big, just think more documentation would be nice so the "Issues" section doesn't get clogged as much.

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -33,6 +33,10 @@ Dependencies
       protobuf-2.5.0
       qrencode-3.4.3
 
+Suggested RAM: 4GB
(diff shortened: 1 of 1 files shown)


prestotron:
Oops, did I end up attempting to merge into master?

I meant to merge into 1.9.

Well, we can worry about this when 1.9 comes :)

langerhans:
Revisit this later, find "sane" numbers maybe

PR #1177, opened 2015-06-23
Rename SendMoneyToDestination to SendMoney.

Sporklin:
Get rid of SendMoney and replace it by the functionality of SendMoneyToDestination.  This cleans up the code, since only SendMoneyToDestination was actually used (SendMoney internally from this routine).


prestotron:
Looks good in my opinion.

PR #1178, opened 2015-06-23
Trigger -alertnotify if network is upgrading without you

Sporklin:
This adds to the "has the rest of the network upgraded to a block.nVersion we don't understand" code so it calls -alertnotify when 51 of the last 100 blocks are up-version. But it only alerts once, not with every subsequent new, upversion block.

And adds a forknotify.py regression test to make sure it works, supported by a -regetest-only undocumented (on purpose, for regression testing only) command-line option -blockversion=N to set block.nVersion.

Tested using forknotify.py:

Before adding CAlert::Notify, get:
Assertion failed: -alertnotify did not warn of up-version blocks

Before adding code to only alert once:
Assertion failed: -alertnotify excessive warning of up-version blocks

After final code in this pull:
Tests successful


prestotron:
Ohhh, I like this :o

PR #1180, opened 2015-06-23
Update the Address Book

prestotron:
I just updated the address book to be as updated as Bitcoins.

It includes a new few things, but should remain the same.

(I had pulled this earlier to the master branch on accident so please ignore that. I have closed it.)


prestotron:
Currently, Travis CI seems to fail. I'm not sure why, maybe by the time someone reads this, it'll be fine, but I tested it, and it worked...

prestotron:
Okay, the scicon.cpp was not needed, so I deleted it, and I had to fix the icons in order for it to build correctly.

Travis CI should be all set now. :)

langerhans:
Thanks, but did you even try to compile your changes locally? First, your filename has a typo, so of course the build fails. Then, you removed the cpp but left in the header and the used methods. That can't work.  

If you want to pull over changes from bitcoin, then you should always cherry pick their changes so the original changes are preserved. Redoing them on your own is prone to errors like this. We follow a very strict process in merging the changes.

At the moment we are working on restricting this process, so you might want to wait with PR'ing changes at this point of time. If you want to help later, please feel free to drop by #dogecoin-dev@freenode so we can talk about how this works out ;)

Closing here for now.

PR #1204, opened 2015-07-23
Add Dogecoin current fee calculation logic

rnicoll:
Introduces 1 COIN/kb fees, rounded up to the next 1 COIN.
Disable free transactions
Dust outputs incur a 1 COIN additional fee.
Add unit tests for fee calculation.
Update existing unit tests with higher transaction values so that transactions are still standard.

### DIFF
--- a/src/dogecoin.cpp
+++ b/src/dogecoin.cpp
@@ -7,6 +7,7 @@
 
 #include "arith_uint256.h"
 #include "dogecoin.h"
+#include "main.h"
 #include "util.h"
 
 int static generateMTRandom(unsigned int s, int range)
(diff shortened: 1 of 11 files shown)


rnicoll:
This isn't fully tested due to time limitations, but I figure it at least moves us closer to a functioning wallet even if we require later cleanup.

langerhans:
Yay for magic numbers. See those that basically prevent sending a normal DOGE tx. Although I think the first one is fine, the other two will complain with 1DOGE fee. I'd suggest 100 DOGE for us? Or 50?

https://github.com/dogecoin/dogecoin/blob/1.10-dev/src/main.cpp#L1035  
https://github.com/dogecoin/dogecoin/blob/1.10-dev/src/qt/walletmodel.cpp#L293  
https://github.com/dogecoin/dogecoin/blob/1.10-dev/src/qt/sendcoinsdialog.cpp#L531

langerhans:
Thanks for the edit. Going to merge after some basic playing with it.

PR #1205, opened 2015-07-26
Use block height 0 when checking for genesis block

rnicoll:
There is no previous block when validating the genesis block as part of contextual checks, so block height cannot be used. Consensus parameters are now taken from height 0 for the genesis check, and from block height only if the it's not the genesis block.


rnicoll:
This is a pre-requisite for #1206, please apply this before attempting to test that PR.

langerhans:
half-tested ACK. I had this pulled into some other testing and it didn't break anything.

PR #1206, opened 2015-07-26
Add Scrypt hash and AuxPoW header to index database

rnicoll:
Add Scrypt hash and AuxPoW header to index database so that index can be validated (a check that both Litecoin and Namecoin disable for simplicity). The index format is incompatible with 1.8, as a result, however, and guidelines on performing the upgrade will need to be prepared.

Given scope of the change, it may be advisable for services to prepare bootstrap files and destroy existing index databases entirely before reindexing. To discuss.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -111,6 +111,9 @@ public:
     //! pointer to the index of some further predecessor of this block
     CBlockIndex* pskip;
 
+    //! pointer to the AuxPoW header, if this block has one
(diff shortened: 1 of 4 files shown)


langerhans:
Well, I'm okay with this. This is also a good reason to have the bip66 changes on 1.8.3 since many miners will probably take their time with 1.10. Merging now.

PR #1207, opened 2015-07-27
Bring client up to date with Bitcoin Core 0.11.0

rnicoll:
Aware we can't commit these yet, but had some time spare and thought I'd get them ready to apply.


langerhans:
utACK. Only touches stuff we didn't really touch yet.

PR #1208, opened 2015-07-27
Remove hard coded switch on whether simplified rewards are used.

rnicoll:


langerhans:
I will merge this now to move forward, even though it misses setting simplified rewards to regtest for easier test vector setup. I'll PR this together with test fixes I'll be working on, cause it _is_ a test fix after all :)

PR #1210, opened 2015-07-28
Minor AuxPoW fixes

rnicoll:
Replaced references to Namecoin with Dogecoin
Set chain ID when constructing a new block to be mined
Added assignment overload so that assigning one BlockVersion to another does a full copy.


rnicoll:
Confirmed working, and testnet blocks mined successfully now.

langerhans:
Tested on regtest too. Looks good, merging.

PR #1211, opened 2015-07-28
Introduce basic Dogecoin branding

rnicoll:
This is intentionally minimal to reduce the number of places it changes, but is sufficient to tell the difference between the client and Bitcoin Core


rnicoll:
Moving to separate branch

PR #1214, opened 2015-07-30
Introduce basic Dogecoin branding

rnicoll:


langerhans:
Looks good here. Merging.

PR #1215, opened 2015-07-30
Add chain height when fetching consensus parameters in initialization

rnicoll:
Minor fix that I accidentally dropped from #1207 :

Add chain height when fetching consensus parameters in initialization, this code was missed in the first patch as it was disabled at the time.


langerhans:
utACK

PR #1217, opened 2015-07-31
Update checkpoints

rnicoll:


langerhans:
Verified them against my local copy of the chain. Looks good.

PR #1218, opened 2015-08-01
Update Bitcoin references and addresses in strings to Dogecoin

rnicoll:
Update untranslated strings to Dogecoin equivalents. Translated strings will be updated in a later separate commit.


rnicoll:
I've added the binary rename commit to this, as the result is then much more coherent (i.e. configuration path and binary names match).

PR #1220, opened 2015-08-02
Revised payment request handling to use genesis block hash

rnicoll:
Revised payment request handling to use genesis block hash instead of network name, enabling
support for more networks that just Bitcoin main and test net.

### DIFF
--- a/src/qt/paymentrequest.proto
+++ b/src/qt/paymentrequest.proto
@@ -1,22 +1,23 @@
 //
-// Simple Bitcoin Payment Protocol messages
+// Simple Dogecoin Payment Protocol messages
+// Derived fromthe Bitcoin Payment Protocol
 //
 // Use fields 100+ for extensions;
-// to avoid conflicts, register extensions at:
-// https://en.bitcoin.it/wiki/Payment_Request
+// to avoid conflicts, register extensions via pull-req at:
+// https://github.com/dogecoin/dips
 //
 
 package payments;
-option java_package = "org.bitcoin.protocols.payments";
+option java_package = "com.dogecoin.protocols.payments";
 option java_outer_classname = "Protos";
 
-// Generalized form of "send payment to this/these bitcoin addresses"
+// Generalized form of "send payment to this/these dogecoin addresses"
 message Output {
-	optional uint64 amount = 1 [default = 0]; // amount is integer-number-of-satoshis
-	required bytes script = 2; // usually one of the standard Script forms
+        optional uint64 amount = 1 [default = 0]; // amount is integer-number-of-satoshis
+        required bytes script = 2; // usually one of the standard Script forms
 }
 message PaymentDetails {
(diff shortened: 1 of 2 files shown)


src/qt/paymentrequest.proto
```diff
@@ -1,22 +1,23 @@
...
 // Use fields 100+ for extensions;
-// to avoid conflicts, register extensions at:
-// https://en.bitcoin.it/wiki/Payment_Request
+// to avoid conflicts, register extensions via pull-req at:
+// https://github.com/dogecoin/dips
 //
 
 package payments;
-option java_package = "org.bitcoin.protocols.payments";
+option java_package = "com.dogecoin.protocols.payments";
```
langerhans: Interesting, I think I never ran protoc on this to actually generate new Protos for dogecoinj. Might be that it won't even be able to generate/use paymentrequests from this code. Should check that for either dogecoinj or altcoinj. Not a merge blocker here tho.

src/qt/paymentserver.cpp
```diff
@@ -770,7 +770,7 @@ void PaymentServer::handlePaymentACK(const QString& paymentACKMsg)
 
 bool PaymentServer::verifyNetwork(const payments::PaymentDetails& requestDetails)
 {
-    bool fVerified = requestDetails.network() == Params().NetworkIDString();
+    bool fVerified = requestDetails.genesis() == Params().hashGenesisBlock;
```
langerhans: I think that should be `HashGenesisBlock()`

src/qt/paymentserver.cpp
```diff
@@ -770,7 +770,7 @@ void PaymentServer::handlePaymentACK(const QString& paymentACKMsg)
...
     if (!fVerified) {
         qWarning() << QString("PaymentServer::%1: Payment request network \"%2\" doesn't match client network \"%3\".")
             .arg(__func__)
```
langerhans: One line below that (github doesn't like me) `.network()` doesn't exist here anymore.

src/qt/paymentserver.cpp
rnicoll: qt/paymentserver.cpp: In static member function ‘static bool PaymentServer::verifyNetwork(const payments::PaymentDetails&)’:
qt/paymentserver.cpp:773:59: error: ‘const class CChainParams’ has no member named ‘HashGenesisBlock’
     bool fVerified = requestDetails.genesis() == Params().HashGenesisBlock();
                                                           ^
qt/paymentserver.cpp:777:56: error: ‘const class payments::PaymentDetails’ has no member named ‘HashGenesisBlock’
             .arg(QString::fromStdString(requestDetails.HashGenesisBlock()))

Going to go with "Looks like no"

rnicoll:
Sorry about that, did a "make distclean" first this time and hopefully that's cleared the compilation error.

langerhans:
Yup, that looks better. Thanks.

PR #1222, opened 2015-08-03
Added wallet_bgcoin.png to QT makefile

rnicoll:


langerhans:
Weird that it worked fine on a local build. But whatever, now that Travis is happy...

Issue #1203, opened 2015-07-13
the gettxoutsetinfo command error


qianshanhai:
I want to fetch the currently mining total, when I use 'gettxoutsetinfo' command, it output:

{
    "height" : 788480,
    "bestblock" : "faf9cde37128227389991653e2911267937e01850f8a443bc6bb7ce6f5fef703",
    "transactions" : 4414277,
    "txouts" : 16550439,
    "bytes_serialized" : 579593118,
    "hash_serialized" : "9a9a3366f9c73d5253a35687bb759c5c33a536fb04c18cbf3c68a1b18cb90c48",
    "total_amount" : -84348075190.47520447
}

Why the total_amount value is not correct?
And how do get this value?
Thanks.

langerhans:
It's counted in an int64_t which won't be able to hold the full amount of currently existing coins (in their smallest unit).  
We could check if we can fix that but it seems to be a small issue since it's just statistics, nothing crucial.

langerhans:
Fixed in #1227

PR #1223, opened 2015-08-03
Changing nTotalAmount of CCoinsStats to uint256 from int64_t

kaykurokawa:
Fix for Issue 1203 (Currently existing coins exceed int64_t)
https://github.com/dogecoin/dogecoin/issues/1203

overzealous size, but uint256 has tests and is commonly used elsewhere in the code

### DIFF
--- a/src/coins.h
+++ b/src/coins.h
@@ -248,7 +248,7 @@ struct CCoinsStats
     uint64_t nTransactionOutputs;
     uint64_t nSerializedSize;
     uint256 hashSerialized;
-    int64_t nTotalAmount;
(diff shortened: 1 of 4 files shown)


langerhans:
Thanks for this! If you would like, could you please resubmit this to the 1.10-dev branch, then we can probably make this available in the next release (after giving is a try of course).  
The master branch is exclusively mutable by release and not pull requests (see the readme).

kaykurokawa:
I see , will do

rnicoll:
We may want to give up on master not being current, BTW, I think it causes more confusion than it saves :-/

PR #1224, opened 2015-08-04
Add paper wallet generator to QT wallet

rnicoll:
I've rebuilt @Andymeows' paper wallet generator around Dogecoin Core 1.10 - this is rough and not yet fully tested, but given the scope of the changes I thought it better to get them in now, than risk further conflicts later.

Testing/code review would be appreciated.


langerhans:
Tested on Debian 8 and prints fine. Tested on Windows 10 and I can't get the "So Print" button to do anything. Have a feeling might be Qt acting up for some reason again.

rnicoll:
Only idea I've got is that for some reason  the print dialog isn't opening. Everything else should spew clear errors. I don't actually have Dogecoin building for Windows yet... err... can you take a look?

langerhans:
Yes, I'll take a closer look. I think this exact issue popped up before and was fixed. I don't remember how and can't find a PR for it, but it will be fixed eventually ;)

rnicoll:
Wonder if #779 is related perhaps...

rnicoll:
http://stackoverflow.com/questions/17175398/deployed-qt5-application-doesnt-print-or-show-print-dialog Suggests it might be missing the minimal library, but checking against #693 all references seem to be added...

rnicoll:
Wait, was reading wrong branch...

langerhans:
Confirmed working on Windows now, thanks! Icon looks nice too:

PR #1225, opened 2015-08-04
Update time-based constants

rnicoll:
This updated various time-based constants to more closely reflect values appropriate to Dogecoin:

* Number of blocks to look back for BIP 66 enforcement is doubled on main net (to about a day and a half's worth, vs 5 days on Bitcoin), and raised to 1,000 on testnet (bringing it up to the same quantity
as Bitcoin testnet).
* BIP 66 enforcement cut-offs are scaled up appropriate.
* Minimum number of blocks and blockchain size are increased for pruning. In comparison to Bitcoin this uses 24 hours as a minimum, rather than 48, although given blocks are rarely full this likely reflects a lot longer in reality.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -37,9 +37,9 @@ public:
 
         // Blocks 0 - 144999 are conventional difficulty calculation
         consensus.nSubsidyHalvingInterval = 100000;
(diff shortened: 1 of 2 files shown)


langerhans:
utACK, the numbers sound sensible.

PR #1226, opened 2015-08-05
Update QT client messages and corresponding translations

rnicoll:
This is a rough first attempt to update QT client messages and translations to Doge
equivalents. Haven't managed to update in Japanese or Chinese translations, and need
to merge in corrections made to the existing translation files as well.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -103,7 +103,7 @@ void setupAddressWidget(QValidatedLineEdit *widget, QWidget *parent)
 #if QT_VERSION >= 0x040700
     // We don't want translators to use own addresses in translations
     // and this is the only place, where this address is supplied.
-    widget->setPlaceholderText(QObject::tr("Enter a Bitcoin address (e.g. %1)").arg("1NS17iag9jJgTHD1VXjvLCEnZuQ3rJDE9L"));
+    widget->setPlaceholderText(QObject::tr("Enter a Dogecoin address (e.g. %1)").arg("DEaT9KZM6b6oDZMr8pj7pWTLZSdtYZFAx8"));
 #endif
     widget->setValidator(new BitcoinAddressEntryValidator(parent));
     widget->setCheckValidator(new BitcoinAddressCheckValidator(parent));
--- a/src/qt/locale/bitcoin_de.ts
+++ b/src/qt/locale/bitcoin_de.ts
@@ -47,31 +47,31 @@
     </message>
     <message>
         <source>Choose the address to send coins to</source>
-        <translation>Wählen Sie die Adresse aus, an die Sie Bitcoins überweisen möchten</translation>
+        <translation>Wählen Sie die Adresse aus, an die Sie Dogecoins überweisen möchten</translation>
     </message>
     <message>
         <source>Choose the address to receive coins with</source>
-        <translation>Wählen Sie die Adresse aus, über die Sie Bitcoins empfangen wollen</translation>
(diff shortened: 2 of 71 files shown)


src/qt/guiutil.cpp
```diff
@@ -103,7 +103,7 @@ void setupAddressWidget(QValidatedLineEdit *widget, QWidget *parent)
 #if QT_VERSION >= 0x040700
     // We don't want translators to use own addresses in translations
     // and this is the only place, where this address is supplied.
-    widget->setPlaceholderText(QObject::tr("Enter a Bitcoin address (e.g. %1)").arg("1NS17iag9jJgTHD1VXjvLCEnZuQ3rJDE9L"));
+    widget->setPlaceholderText(QObject::tr("Enter a Dogecoin address (e.g. %1)").arg("1NS17iag9jJgTHD1VXjvLCEnZuQ3rJDE9L"));
```
langerhans: Address needs change. Should use the same example address that made it in in the other PR already.

src/qt/locale/bitcoin_de.ts
```diff
@@ -3170,15 +3170,15 @@ Beispiel: alertnotify=echo %%s | mail -s "Bitcoin Alert" admin@foo.com
     </message>
     <message>
         <source>Copyright (C) 2009-%i The Bitcoin Core Developers</source>
-        <translation>Urheberrecht (C) 2009-%i Die "Bitcoin Core"-Entwickler</translation>
+        <translation>Urheberrecht (C) 2009-%i Die "Dogecoin Core"-Entwickler</translation>
```
langerhans: Current one says  

Copyright © 2009-2014 The Bitcoin Core developers
Copyright © 2013-2014 The Dogecoin developers

src/qt/locale/bitcoin_de.ts
```diff
@@ -3234,7 +3234,7 @@ Beispiel: alertnotify=echo %%s | mail -s "Bitcoin Alert" admin@foo.com
     </message>
     <message>
         <source>RPC SSL options: (see the Bitcoin Wiki for SSL setup instructions)</source>
-        <translation>RPC-SSL-Optionen (siehe Bitcoin-Wiki für SSL-Einrichtung):</translation>
+        <translation>RPC-SSL-Optionen (siehe Dogecoin-Wiki für SSL-Einrichtung):</translation>
```
langerhans: I think we left that at Bitcoin-Wiki since the other doesn't even have that info (if it even still exists).

src/qt/locale/bitcoin_de.ts
rnicoll: Yeah, we may have to add a new translation to the files, unfortunately...

rnicoll:
Haven't solved the copyright issue (just reverted it to "Bitcoin Core"), as we'll need to add a new translation and modify the UI. Otherwise it's in a lot better state. Could you check the merged translations and tell me if any more need culling? I've cleared the worst messes out, at least.

langerhans:
Still feels like a lot of unneeded changes there. I looked at master and collected actually contributed file updates. There would be zh_CN, zh_TW, it, nl and ko_KR.  
For Japanese a find&replace from `ビット` to `ドージ` should work as far as I see from the Android Wallet, where a reddit user gave me the hint a while ago.

rnicoll:
Okay, try that?

langerhans:
That looks a lot better to me. Ignore my comment about Japanese for now as it turns up with weird results on Google, although the one that's bitcoin at least resembles the meaning a lot better :D  
I still want to compile this once and look over it, but so far ACK.

PR #1237, opened 2015-08-10
Fixed Windows and Mac setup scripts.

Sporklin:
<3


langerhans:
Windows confirmed building correctly on gitian with that. Mac needs more work, but I'll get to that. Thanks!

PR #1241, opened 2015-08-12
qt: Introduce PlatformStyle

rnicoll:
Introduce a PlatformStyle to handle platform-specific customization of the UI.

This replaces 'scicon', as well as #ifdefs to determine whether to place icons on buttons.

The selected PlatformStyle defaults to the platform that the application was compiled on, but can be overridden from the command line with `-uiplatform=<x>`.


langerhans:
Tested all three settings and no params. Looks fine here.

PR #1243, opened 2015-08-13
Replace Bitcoin with Dogecoin in tooltip

langerhans:
Saw it while testing


rnicoll:
Think these were in #1241 and just not highlighted, apologies.

langerhans:
Uhh, I'll check it out tomorrow-ish.

langerhans:
Rebased and changed the one occurance left.

PR #1244, opened 2015-08-13
Add dogechain and chain.so as default block explorers

rnicoll:


langerhans:
I guess this needs rebase on the rebase :D

rnicoll:
Re-re-based

PR #1252, opened 2015-08-15
Disable AuxPoW mining in the beta client

rnicoll:


langerhans:
utACK, mining on the first beta is not a good idea, agreed. Waiting for the patch by @patricklodder so we can focus on testing this part more until next beta.

PR #1261, opened 2015-08-22
Update main.cpp

ChrisTriesToDoBest:


langerhans:
Uhm.... Nope

rnicoll:
To expand on that:

1. This causes an uncontrolled hard fork as soon as it's deployed.
2. There's no analysis of why this is a positive improvement.
3. There's no clear rationale for miners to adopt this change.

Essentially this appears to badly break consensus, likely leaving the resulting fork extremely vulnerable to 51% attacks (or given likely mining power vs that of main chain, 99.9% attacks)

PR #1284, opened 2015-09-11
Add Dogecoin copyright to about window

rnicoll:
Alternative to #1283 - if 1283 renders correct, please use it, I've had to put two new lines in between the copyright notices to get them to format correctly :-/


rnicoll:
Closing, will raises as a PR against #1283

PR #1290, opened 2015-10-02
Change default soft block size limit to hard max

peoplma:
Fairly straight forward.  Changes the default configuration of a mining node from mining 750kB blocks to mining the max of 1MB blocks.  Pros:  During a hypothetical spam attack miners would be set up by default to handle 25% more transactions, would clear mempool that much faster, would cost that much more in fees to cause disruption to regular users.  Cons:  More hard drive space and bandwidth would be required, possibly leading to higher orphan rates during spam attack.  Ultimately though, it's just a default setting and miners are free to adjust their own personal soft-limit down to whatever size blocks they want to mine.

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -48,7 +48,7 @@ class CValidationState;
 struct CNodeStateStats;
 
 /** Default for -blockmaxsize and -blockminsize, which control the range of sizes the mining code will create **/
-static const unsigned int DEFAULT_BLOCK_MAX_SIZE = 750000;
+static const unsigned int DEFAULT_BLOCK_MAX_SIZE = 1000000;
 static const unsigned int DEFAULT_BLOCK_MIN_SIZE = 0;
 /** Default for -blockprioritysize, maximum space for zero/low-fee transactions **/
 static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = 27000;


patricklodder:
I think this should be pulled against 1.10-dev instead, @langerhans ?

langerhans:
Technically it should, but on the other hand I don't quite see the need for this change. Have a look at the results of the stress test end of August. These are the recorded block sizes:
testnet-blocks

With these transaction rates:
testnet-transactions

Even the highest rates produced blocks of around half the current soft limit. Of course could a larger spam attack fill up more block space, but at the point where someone fills 750kb, they won't care much for the other 250kb.
Or am I missing something?

peoplma:
After I submitted I thought to myself "that should probably have gone into -dev" haha apologies.  Thanks for the graphs @langerhans.  You're right it's not a big deal either way, I just never really understood the logic behind why the soft limit default was set lower than the hard limit since if miners are having trouble with orphans they can lower their own personal custom limit.  At least one block during the test pushed closed to 750kB https://chain.so/block/DOGETEST/5323361ea951436ded2b4c755d4aa765d7dd60fd30ea2b3ddf2bfb8c3fecb817.  250kB more is 250kB more.

patricklodder:
> Or am I missing something? 

testnet has the difficulty reset so from a block perspective it's not comparable with livenet.

peoplma:
Also the assumption is that an attacker would max out the capacity and then some, this change would get us back to normal faster after the attack stopped.

patricklodder:
@peoplma that's correct, my thoughts on that:

1. 1MB blocks will take you time to parse, especially when the utxo are all over the place and there are tons of operations to evaluate (think p2sh with custom redeemscripts, we're opening up for that currently), will take a huge toll on dogecoind performance. I know of some infra that had to scale up during the bitcoin spam, because some blocks were taking 40-50 seconds to be parsed. Imagine that on dogecoin, with 1-minute blocks.
2. We already have 7.5 times the capacity of bitcoin. we can increase it to 10 times but honestly, if someone can plan a consistent attack that does something like 75 times the capacity bitcoin has, then it's not too hard to scale that to 100 times either. I'd think the attack will probably just scale with the limit.
3. The difference in min fee cost will rise from 750 doge per minute to 1000 doge per minute, so that's a change from approx USD 0.09 to USD 0.12/minute right now, I'm not sure if that makes any difference, probably not even at 10 times minimum fee, with 10 times max capacity: the cost would go from $9 to $12 per minute, with the total subsidy for merge-mining ltc+doge currently around $30 a minute.
4. The threat is real in the form that it can be done. I can do anything between 30 and 3000 tps given i have enough funds to pay the fees and stay above dust outputs, thus so can any other fool. I'm not special. What's there to gain from such an attack though? What's the incentive?

My bottom line: I see no real reason to increase the limit right now, nor do I see real good reasons not to. The change will imho not help much in case of an attack. But that's just my opinion right now, I haven't seen a really good argument yet, maybe one will come? :)

peoplma:
Yep, I think you pretty much nailed it Patrick. 1MB blocks every minute will certainly be intensive on the CPU, but I think it would be more RAM limited than CPU limited, and 1MB blocks will clear the mempool faster than 750kB blocks will.  But yes, I agree this isn't a solution to the potential attack, but it does make it 25% harder, and as you say there's no real good reason not to.

Another potential solution for the mempool would be what bitcoin XT did which is (by default) start randomly dropping transactions when mempool reaches 50,000 (user configurable variable).  But that's another discussion :)

patricklodder:
If you're worried about memory, then this is not the fix you're looking for, nor if you're worried about cpu. The only thing this fixes is fee market in case the blocks get full, and the only way to combat it imho is up the fees pre-emptively. How does 200 doge per kb sound? We'd be close to btc fees then :)

peoplma:
Nooo haha I keep seeing Ross talking about increasing fees.  Doge won't be an easy to use starter crypto anymore!  Better to keep 1 doge per kB and risk getting hit by spam every now and then than to make doge unusable for newbies to play around with.

patricklodder:
Figured as much. I'm NAK'ing this for 1.10.0 (and it's on the wrong branch) because I feel we're too late in the process to take this change in anyway. If others want this for 1.10.1, I won't NAK it, but I am not convinced we need it, I see more cons than pros.

As for the XT patch that you mentioned (and perhaps others too, haven't looked in-depth) I think we can look at those for 1.10.1 too, on a case-by-case basis. Better mempool management may seem like a luxury right now, but might be worth to consider, imho.

Sorry if this is disappointing.

langerhans:
I basically agree with the statement that there aren't any really good arguments for either side.  
So I'm NACK'ing this change. As is said, this value can be changed by miners anyway. Also scaling this attack from 750kb to 1000kb doesn't really make much of a diff so it won't help here.

patricklodder:
PS:
>  It would make accepting 0 confirmation transactions substantially more risky though.

NEVER, EVER, ACCEPT 0-CONF DOGECOIN TRANSACTIONS. We have fast block timing, use it to your advantage.

peoplma:
Yes yes I know Patrick haha.  Anyway thanks for the thoughtful discussion and consideration guys :)

patricklodder:
Closing this here. If anyone wants to further discuss block size, please open an issue :)

PR #1312, opened 2015-10-20
Minor changes

leuqarte:
Light cleanup


langerhans:
Thanks for this. Would it be possible to squash the three commits into one so we have a cleaner commit history? Thanks :)

leuqarte:
still new to the way GitHub works, sorry :D

rnicoll:
Thanks!

PR #1315, opened 2015-11-16
1.10.2 batch1

rnicoll:
This merges in patches up to c09dfd06fc2d183b996503d29da502ab0b49cb26 from the 1.10-dev branch. These have been rebased to try cleaning up the commit set, so are not 1:1 identical to the original patches.


rnicoll:
Applied some later patches as fixups, now this is up to date as far as d62c7aa09ce9a3de1e8f45d36e752b79566ad47c on the 1.10-dev branch.

PR #1322, opened 2016-01-25
Replaced bitcoin mention to dogecoin.

Sporklin:
It was already changed in the translation files, however this was somehow missed.


rnicoll:
Tested, thanks for fixing this

PR #1374, opened 2017-05-31
Update tidy_datadir.sh

just-an-dev:
replace Bitcoin to Dogecoin


langerhans:
This seems not required as Dogecoin was always based on Bitcoin > 0.8. So this file has no real use for any Dogecoin user. Also this is the wrong branch for PRs, they should go to the most current -dev branch. Thank you anyway tho :)

PR #1378, opened 2017-08-09
Update build-unix.md

katin:
libdb is currently version 5.3
Not sure what the new Oracle download link should be, though.
Maybe http://download.oracle.com/otn/berkeley-db/db-6.2.32.NC.tar.gz ?

Ref:
http://www.oracle.com/technetwork/database/database-technologies/berkeleydb/downloads/index.html

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -61,11 +61,11 @@ Build requirements:
 	
 for Ubuntu 12.04 and later or Debian 7 and later, libboost-all-dev has to be installed:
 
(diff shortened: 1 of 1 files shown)


katin:
I see the configure script specifically looks for 5.1... I'm guessing there's something in 5.3 that isn't compatible. (I saw some reference to mobile wallets?)
I'll see if I can compile instructions to find and install the 5.1 version, as they are no longer available via apt-get repos.

PR #1379, opened 2017-08-13
Merge in basics of AuxPoW support

rnicoll:


rnicoll:
This isn't completed (currently it won't sync past the first hard fork anyway), but getting the skeleton of this code in now. If you can see any obvious mistakes please let me know.

Also we need to pay the Namecoin devs when 1.14 ships, because honestly this is mostly their code.

qa/rpc-tests/rest.py
```diff
@@ -219,24 +223,37 @@ def run_test(self):
...
+        assert_greater_than(headerLen, 80)
         response_header_str = response_header.read()
-        assert_equal(response_str[0:80], response_header_str)
+        assert_equal(response_str[0:headerLen], response_header_str)
 
         # check block hex format
         response_hex = http_get_call(url.hostname, url.port, '/rest/block/'+bb_hash+self.FORMAT_SEPARATOR+"hex", True)
         assert_equal(response_hex.status, 200)
         assert_greater_than(int(response_hex.getheader('content-length')), 160)
+<<<<<<< HEAD
```
langerhans: Merge errors here and below

langerhans:
Thanks! I guess after that we need to be looking at getting DigiShield and all the other crap in to get some syncing past all the forks going. There is also the issue of BIP9 and if we want to fork to support it but I'm not too keen on planning a fork right now. Maybe do that in the next version and shove segwit in too cause why the hell not ;)

rnicoll:
I think all of the code is in, it just doesn't so much "work". Getting it to sync and fixing the tests are next tasks. I'll fix the merge stuff in a minute.

rnicoll:
On forking/Segwit; I'm a fan of 1.15 being the BIP 9 & Segwit release. There's other stuff I'd like in a fork, but we can discuss that later on.

PR #1385, opened 2017-09-23
Sync mining code from Namecore to resolve unit test failures

rnicoll:
Fixes a bunch of test failures.

Four tests still fail, I think the auxpow tests are failing because the behaviour has changed to reject blocks without auxpow detail, rather than deferring to normal PoW checks. Need to verify.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -128,7 +128,7 @@ void BlockAssembler::resetBlock()
     blockFinished = false;
 }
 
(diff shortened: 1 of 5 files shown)


rnicoll:
Have rebuilt, and now realised I think the tests are incorrect. Certainly the one at line 369 of auxpow_tests.cpp seems to expect a valid block to be rejected for uncertain reasons.

PR #1392, opened 2017-11-19
Introduce Digishield retarget code to 1.14

rnicoll:


rnicoll:
Next stage is to correct v3 & v4 enforcement heights, see if AuxPoW works, and merge in the proper chainparams resolution code.

PR #1397, opened 2017-12-09
Added RPC throw

hswick:
Hi! I made a small change to rpcserver to try to get my feet wet with contributing. This change is based on this commit from 1.10 https://github.com/dogecoin/dogecoin/commit/8d5284422914164f73bfbdf3b275879da355af30

I've checked that it compiles, the rpc_tests don't pass but I believe it is related to aux_pow. Would love feedback if this is useful or not!

To the moon!


langerhans:
Hey sorry I'm getting back to you so late about this, been a busy few days...
Anbyway, thank you for the contribution. Our current plan is to get the fundamentals in before we start pulling in "the rest", however much that might be. I'll keep this here open. As soon as we have a good idea of which commits we require help with, we'll try to list those somewhere for contributors to pick. That should make the process easier for all of us :)

hswick:
Sweet, sounds good! To the moon

rnicoll:
You're going to hate me for this... doesn't the line `MoneyRange(amount)` from upstream now replace this?

langerhans:
Yeah this has been fixed from upstream already. Sorry.

PR #1403, opened 2017-12-26
Change BIP65/66 enforcement to match Dogecoin

rnicoll:
Introduce estimated BIP66 enforcement height (TBC), and switch BIP65 enforcement to supermajority rather than height based as Dogecoin doesn't have it yet.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -71,7 +71,9 @@ public:
     CMainParams() {
         strNetworkID = "main";
         consensus.nSubsidyHalvingInterval = 100000;
-        // XXX: BIP heights and hashes all need to be updated to Dogecoin values
+        consensus.nMajorityEnforceBlockUpgrade = 1500;
+        consensus.nMajorityRejectBlockOutdated = 1900;
+        consensus.nMajorityWindow = 2000;
         consensus.BIP34Height = 139314;
         consensus.BIP34Hash = uint256S("0x2c23095473d1afabfb630d41a395364e69da264a73c04dbcb4756c7d36c9b994");
         // consensus.BIP65Height = 1032483; // Not enabled in Doge yet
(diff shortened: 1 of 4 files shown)


src/chainparams.cpp
```diff
@@ -182,11 +185,14 @@ class CTestNetParams : public CChainParams {
     CTestNetParams() {
         strNetworkID = "test";
         consensus.nSubsidyHalvingInterval = 100000;
-        // XXX: BIP heights and hashes all need to be updated to Dogecoin values
-        consensus.BIP34Height = 21111;
-        consensus.BIP34Hash = uint256S("0x0000000023b3a96d3484e5abb3755c413e7d41500f8e2a5c3f0dd01299cd8ef8");
-        consensus.BIP65Height = 581885; // 00000000007f6655f22f98e72ed80d8b06dc761d5da09df0fa1dc4be4f861eb6
-        consensus.BIP66Height = 330776; // 000000002104c8c45e99a8853285a3b592602a3ccde2b832481da85e9e4ba182
+        consensus.nMajorityEnforceBlockUpgrade = 1500;
+        consensus.nMajorityRejectBlockOutdated = 1900;
```
langerhans: On master those are 501 and 750.

src/chainparams.cpp
```diff
@@ -182,11 +185,14 @@ class CTestNetParams : public CChainParams {
...
+        consensus.nMajorityWindow = 2000;
```
langerhans: Also this is 1000.

src/validation.cpp
```diff
@@ -3022,11 +3028,15 @@ bool ContextualCheckBlockHeader(const CBlockHeader& block, CValidationState& sta
...
     // check for version 2, 3 and 4 upgrades
     // Dogecoin: Version 2 enforcement was never used
-    if((block.GetBaseVersion() < 3 && nHeight >= consensusParams.BIP66Height) ||
-       (block.GetBaseVersion() < 4 && nHeight >= consensusParams.BIP65Height))
+    if((block.GetBaseVersion() < 3 && nHeight >= consensusParams.BIP66Height))
             return state.Invalid(false, REJECT_OBSOLETE, strprintf("bad-version(0x%08x)", block.GetBaseVersion()),
                                  strprintf("rejected nVersion=0x%08x block", block.GetBaseVersion()));
 
+    // Dogecoin: Introduce supermajority rules for v4 blocks
+    if (block.nVersion < 4 && IsSuperMajority(4, pindexPrev, consensusParams.nMajorityRejectBlockOutdated, consensusParams))
```
langerhans: Shouldn't this use `block.GetBaseVersion()`?

rnicoll: Good question - master doesn't https://github.com/dogecoin/dogecoin/blob/master/src/main.cpp . Going to change that, but suspect we haven't been enforcing v3 properly from the looks of things and were saved by AuxPoW.

langerhans: Yeah, master doesn't use them in that function at all 🤔

rnicoll:
Okay, up for re-review

PR #1405, opened 2017-12-26
Set SegWit timeout to zero to disable it

rnicoll:
This is how SegWit is actually meant to be disabled, rather than the weird hack I put into mining.


rnicoll:
Generally keeping changes to a minimum, the timeout is enough to disable it, so going with that for now. Mostly the fewer changes, the easier when we rebase all of this for 1.15 :)

PR #1406, opened 2017-12-26
Verify when doing 'backupwallet' that destination is not the same path of the original wallet.dat file.

### LINKED ISSUE #1402: Backupwallet RPC call destroyed wallet if target equals the source wallet file
See this issue in bitcoin : https://github.com/bitcoin/bitcoin/issues/11375

Actually happened to me today.

rllola:
Fix #1402


rnicoll:
Looks good from a quick glance, thanks for this! I'll give it a quick test before merging, but should go in this week.

rnicoll:
Many thanks for this!

PR #1411, opened 2017-12-30
1.14 check block version

rnicoll:


rnicoll:
This stops the warning about block version numbers.

src/validation.cpp
```diff
@@ -1721,12 +1721,14 @@ int32_t ComputeBlockVersion(const CBlockIndex* pindexPrev, const Consensus::Para
...
+    // FIXME: Active version bits after the always-auxpow fork!
+    /* for (int i = 0; i < (int)Consensus::MAX_VERSION_BITS_DEPLOYMENTS; i++) {
         ThresholdState state = VersionBitsState(pindexPrev, params, (Consensus::DeploymentPos)i, versionbitscache);
         if (state == THRESHOLD_LOCKED_IN || state == THRESHOLD_STARTED) {
             nVersion |= VersionBitsMask(params, (Consensus::DeploymentPos)i);
         }
-    }
+    } */
 
     return nVersion;
```
langerhans: Wouldn't this now always return `VERSIONBITS_TOP_BITS`?

rnicoll:
Reverted the block number calculation change and just jammed the version number in as fixed again.

PR #1413, opened 2018-01-01
Add Dogecoin current fee calculation logic

rnicoll:


rnicoll:
Early tests suggest this is producing crazy high fees (i.e. https://chain.so/tx/DOGETEST/5280a8c612ba0502c2b024fcaeaa37abf0eaee42abe4e9ff95ad109639f4b484 ) but at least it's closer than not having it in, merging for now

PR #1414, opened 2018-01-01
Change result from getauxblock to BIP22 validation state

rnicoll:


langerhans:
This should probably include https://github.com/dogecoin/dogecoin/commit/e2334c99e

rnicoll:
Actually with https://github.com/dogecoin/dogecoin/commit/e2334c99e it's basically no-change isn't it? This adds a more complex return which `e2334c99e` then turns back into a simple one?

This is probably a "Does not need to be applied" in hindsight.

langerhans:
https://github.com/dogecoin/dogecoin/commit/e2334c99e also adds a compatibility method though. I do not know what current pool implementations use so I'd rather have us offer both variants. Or am I missing something here?

rnicoll:
I'll check, but it doesn't look like `getauxblockbip22` is actually exposed anywhere

langerhans:
`getauxblock` decides based on the parameters what to return. So it is kinda exposed.

rnicoll:
Ah, got it! Yup will get that glued on in.

PR #1415, opened 2018-01-01
Change count type used by `gettxoutsetinfo`

rnicoll:
changing CAmount (is a int64_t) to arith_uint256 for nTotalAmount in CCoinsStats to prevent overflow

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -808,7 +808,7 @@ struct CCoinsStats
     uint64_t nTransactionOutputs;
     uint64_t nSerializedSize;
     uint256 hashSerialized;
-    CAmount nTotalAmount;
+    arith_uint256 nTotalAmount;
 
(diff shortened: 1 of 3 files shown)


rnicoll:
Result looks sane, but I still need to check total against a 1.10 node before we merge this.

rnicoll:
This produces slightly different results on 1.10 and 1.14, but personally I'm fairly comfortable that's because we're not doing ill-advised things with double precision numbers, rather than a fault in the new version.

rnicoll:
1.14: `"total_amount": 104591773511.59000089`
1.10: `"total_amount": 104592845336.59001160`

langerhans:
Did you compare the value of `txouts`? The UTXO set changes quickly might just be that the two nodes were in different states.

rnicoll:
I did, and compared on different blocks as well, the error between the two is fairly consistent. As said, fairly confident this is due to misuse of doubles in 1.10.

PR #1417, opened 2018-01-02
Rebase 1.14 work on 1.15 so basic sync completes

rnicoll:


rnicoll:
I definitely want to release 1.14 as its own version, to avoid problems with deprecated/removed RPC functions, but rebasing that work on 1.15 now so we can test the two in parallel.

Most importantly, note that right now this doesn't include `getauxpow` because neither does Namecore 0.15.

rnicoll:
Abandoning the idea of staggered PRs for 1.15, and going to make a single big-bang PR. Closing this one .

PR #1420, opened 2018-01-05
Change fPIE to fPIC

D3m0nKingx:
By building with fPIE, the resulting binaries are "LSB shared objects" not LSB executables. When trying to run the program (double-click) with mouse it doesn't start. Therefore this change fixes this.


rnicoll:
I believe this is a change in Qt5 since 1.10 went out, I've seen something similar in 1.14 and presumed it was me, but merging as it seems more widespread.

rnicoll:
Aaaaand I've merged into master when it should be 1.10-dev. I'll get that fixed.

PR #1425, opened 2018-01-08
Bring 1.14 fees in line with 1.10

rnicoll:
Reduce fee defaults to saner amounts

### DIFF
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -49,7 +49,7 @@ static const CAmount DEFAULT_TRANSACTION_FEE = 0;
 //! -fallbackfee default
 static const CAmount DEFAULT_FALLBACK_FEE = COIN;
 //! -mintxfee default
-static const CAmount DEFAULT_TRANSACTION_MINFEE = COIN / 10;
+static const CAmount DEFAULT_TRANSACTION_MINFEE = COIN;
 //! minimum recommended increment for BIP 125 replacement txs
 static const CAmount WALLET_INCREMENTAL_RELAY_FEE = COIN * 5;
 //! target minimum change amount
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -67,7 +67,7 @@ static const unsigned int DEFAULT_TX_CONFIRM_TARGET = 6;
 //! -walletrbf default
 static const bool DEFAULT_WALLET_RBF = false;
 //! Largest (in bytes) free transaction we're willing to create
-static const unsigned int MAX_FREE_TRANSACTION_CREATE_SIZE = 1000;
+static const unsigned int MAX_FREE_TRANSACTION_CREATE_SIZE = 0;
 static const bool DEFAULT_WALLETBROADCAST = true;
 static const bool DEFAULT_DISABLE_WALLET = false;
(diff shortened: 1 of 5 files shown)


src/wallet/wallet.h
```diff
@@ -47,9 +47,9 @@ static const unsigned int DEFAULT_KEYPOOL_SIZE = 100;
 //! -paytxfee default
 static const CAmount DEFAULT_TRANSACTION_FEE = 0;
 //! -fallbackfee default
-static const CAmount DEFAULT_FALLBACK_FEE = COIN * 20;
+static const CAmount DEFAULT_FALLBACK_FEE = COIN;
 //! -mintxfee default
-static const CAmount DEFAULT_TRANSACTION_MINFEE = COIN;
+static const CAmount DEFAULT_TRANSACTION_MINFEE = COIN / 10;
```
langerhans: Why ist this `COIN / 10`? From 1.10: `CFeeRate minRelayTxFee = CFeeRate(COIN);
`

rnicoll: The idea is to intentionally loosen the minimum so we've got some scope for the price to continue upwards without causing problems.

langerhans: Given we are always creating transactions at `minRelayTxFee` this change would be effectively reducing the fee for every Core user to 0.1 DOGE. I think you meant to change `DEFAULT_MIN_RELAY_TX_FEE`?

Turupawn:
I don't see a reason to reduce the fee. 1Ð fee is easy to understand and explain to my fellow friends new to the cryptos.
I don't think using decimals (0.1Ð) is making doge easier to understand.
I understand that the dogecoin it's rasing it's value compared to usd.
But last time I checked 1Ð = 1Ð

rnicoll:
So the default amounts will stay the same, so 1 DOGE will be the normal fee, this is just giving us scope for moving to a more flexible model later. Realistically we have to avoid turning into something that's too valuable to use, it would be crazy.

GaryLittlemore:
At the current price $1 is 0.01475808492078891 doge, keeping the fee low is a must. Could it not be coded that the fee would be variable depending on the price but it would always be  equivalent to say 50 Cent?

langerhans:
No, that is not possible as this would require the client knowing about exchange rates. Ain't gonna happen ;) Fees can already be set by users and miners. These are merely default fees. Yet, I'd be fine with keeping them at 1 DOGE as this is just over a penny. Even a tenfold increase in price would make it 10-15 cents which is still a lot better than paying a 50$ fee on a 25$ transactions cause you have some small outputs to spend 😂

PeterOdeus:
The 1-doge fee is a cognitive cornerstone for simplicity. If we're
anticipating a price-increase as a general tidal effect of all crypto, I'd
rather contemplate an event like "The Doubling": A 2:1 increase of both
mining fee and wallet holdings. Effectively halving the Dogecoin price
while maintaining the marketcap.

rnicoll:
I've just realised why everyone's commenting on this...

Shibes, I screwed up the first version of 1.14 and it had 10x regular fees. This was just me putting it back down to 1 DOGE. Stop panicking!

PR #1427, opened 2018-01-08
Enforce minor Dogecoin parameters

rnicoll:
* Enable block constraints properly to reflect that Doge always had BIP16 and BIP30
* Update time-based constants including prune block size and fork detection parameters


Dafeesh:
I believe hard fork conditions were kept in for historical (or theoretical replay purposes). At least that's the reason why Bitcoin leaves them in. Should Dogecoin keep them in, too? I'm honestly not sure.

rnicoll:
You mean the BIP16/BIP30 conditions? I'm leaving the code in but disabled because it makes it easier to re-apply the patches for later versions.

PR #1430, opened 2018-01-09
Add block height to block notification substitution options

rnicoll:


greenbigfrog:
What use cases are there for this?

rnicoll:
It's backwards compatible with previous releases of Dogecoin Core. I'm not strictly speaking sure who actually uses it, but we'd rather not drop functionality we had unless it causes problems.

greenbigfrog:
I've never seen the %i option, haven't really used the notification either way in a while, so I must have just missed it.

PR #1432, opened 2018-01-09
Paper wallet creator for Dogecoin QT

rnicoll:


langerhans:
I guess this needs a rebase after #1429 is in.

PR #1433, opened 2018-01-09
Update payment protocol to match Dogecoin

rnicoll:


src/qt/paymentrequest.proto
```diff
@@ -1,24 +1,25 @@
 //
-// Simple Bitcoin Payment Protocol messages
+// Simple Dogecoin Payment Protocol messages
+// Derived fromthe Bitcoin Payment Protocol
```
greenbigfrog: **from the**

Issue #1442, opened 2018-01-10
getheaders not serializing correctly in 1.14


rnicoll:
Seeing an issue with `1.14-dev` currently where a remote sendheaders command can trigger a client crash. It looks like the headers are being loaded from disk without their AuxPoW section, so when the client tries sending it back, serialization fails and it crashes.

Current example seems to be testnet block `1561d36f48fbc6799d803b94768e317c25b5aec6150c61ef7276f317af9ecf47` but currently confirming if it's always that block, or just random chance here.

langerhans:
Is this still an issue?

PR #1443, opened 2018-01-10
Load AuxPoW blocks from index correctly

rnicoll:
Load AuxPoW blocks from index correctly, by setting the version on the block before trying to check if it's AuxPoW. Previously the AuxPoW part of the block was never loaded because the check always failed.


rnicoll:
This should address #1442 although I don't have an easy reproduction case, so hard to be entirely sure.

PR #1464, opened 2018-01-27
Update release notes for 1.14

rnicoll:


Bestbitcoinmanager:
https://t.me/DinoParkBot?start=447783279

doc/release-notes.md
```diff
@@ -1,33 +1,45 @@
 (note: this is a temporary file, to be added-to by anybody, and moved to
 release-notes at release time)
 
-Bitcoin Core version 0.14.0 is now available from:
+Dogecoin Core version 0.14.0 is now available from:
```
langerhans: 1.14

doc/release-notes.md
```diff
@@ -1,33 +1,45 @@
...
 
-  <https://bitcoin.org/bin/bitcoin-core-0.14.0/>
+  <https://dogecoin.com/bin/dogecoin-core-0.14.0/>
```
langerhans: That URL is probably wrong. Should point to GitHub maybe?

rnicoll: I've rewritten it to point to nowhere that works in Github, but mostly it's a placeholder until we get to the release itself.

PR #1465, opened 2018-01-28
Clean up RPC tests

rnicoll:
* Switch to Python 3.5 as it's the default on Ubuntu LTS
* Enable `p2p-fullblock` tests as they should work (but they don't, and that's a concern)
* Scale up transaction amounts in `fundrawtransaction` to try resolving the failure when funding from a watch only address

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -103,7 +103,7 @@ testScripts = [
     'wallet-hd.py',
     'walletbackup.py',
     # vv Tests less than 5m vv
-    # 'p2p-fullblocktest.py',
+    'p2p-fullblocktest.py',
     'fundrawtransaction.py',
     #'p2p-compactblocks.py',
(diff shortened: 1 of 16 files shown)


rnicoll:
Concerned that `fundrawtransaction` definitely seems unwilling to use watch only addresses, and I can't see any clear reason why. I _think_ it may not be recording the payment to the watch only address correctly.

This is the test at line 617 of `fundrawtransaction.py`.

rnicoll:
`fundrawtransaction` now gets as far as the TX fee tests, which break because of our funding schedule, at line 668

rnicoll:
I've cleaned up a bunch of the code for coinbase maturity checks, but it looks like they're genuinely failing. It's as if somehow it can't find the regtest parameters at all. Even more weirdly, it should return the AuxPoW parameters (effective height 20), yet it actually returns the baseline ones (effective height 0).

PR #1470, opened 2018-02-04
Re-introduce alert functionality

rnicoll:


rnicoll:
Opening this early to show work in progress, please use this for any further work. Need to:

* Critical: Process for sending out messages has changed, and it looks like we should trigger a signal that the net processing layer then creates messages from. Have a look at the `UpdatedBlockTip()` signal for an example of how I think this should now work.
* Check whether broadcast of alerts on connect is done correctly
* Check I've got the correct keys in place
* Someone with the actual signing key needs to verify this on testnet

donpdonp:
apologies if this is the wrong place to ask, but if i remember correctly, bitcoin dropped alerting becuase the 'one true alert signing key' was too much of a single point of failure. a lot of mayhem could be had if that key were ever compromised. what is the motivation for bringing back alerts in dogecoin?

rnicoll:
It's important to understand that it's a risk/reward ratio. Bitcoin is now well known enough that in case of a major security issue that requires urgent patching, they call up some journalists and it becomes a news article broadcast globally.

Dogecoin doesn't have that luxury, so we need to retain the functionality for now.

langerhans:
Confirmed working on 1.10 and 1.14 by sending an alert on testnet. They showed up in both clients nicely.

donpdonp:
it might be useful to notify the doge chain sites that this exists - they might want to add it to their UI.

rnicoll:
It's part of 1.10, but I'll add notes in 1.14 to make sure they realise it's remained in the code rather than staying synced with Bitcoin's changes.

PR #1474, opened 2018-02-18
Update tx output in p2p-fullblocktest.py

rnicoll:
* Update tx output in p2p-fullblocktest.py to match Dogecoin coin reward.
* Disable the re-org test as it disconnects the mini-node before the re-org completes.


rnicoll:
This leaves two failing tests we don't have good reasons to disable:

* `p2p-fullblocktest` re-org (diasbled because it drops the Python mininode, which likely reflects changes required to the test rather than the node)
* `mempool_limit`, where the old transaction isn't dropped from the pool. I think this is tx fee related, but not sure exactly how.

I'd not consider either blocking at this point, although the latter likely indicates reduced effectiveness of spam filters.

mspaansen:
😀

PR #1485, opened 2018-04-22
Create v1 transactions for compatibility with 1.10

rnicoll:
Create v1 transactions for compatibility with 1.10. See https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki for context.

This should NOT be merged into 1.16.


Fulvio55:
Rats.... There goes my patch!

Hehe. :)

PR #1498, opened 2018-05-23
Update README.md

Computer-Genuis:
some places refer to doge as xdg


Sporklin:
Our four letter DOGE is standard to most exchanges for listing purposes to avoid this exact issue. 

It was until recently the unofficial ISO code relating until it was found to be conflicting there with existing connections. References to XDG in any do not exist in any official manner as the currency abbreviation exists in the space already which is not attached to Dogecoin. This is part of why before the standard moved to four letters, DogeCoin was able to be listed as DOGE in majority of usage outlets.

This is perhaps why it has never been added, and why I do not think it should be.

Computer-Genuis:
Why not add a fast facts section? Also, why not also put dogecoin in the brackets?

Computer-Genuis:
I changed it

Computer-Genuis:
Many newcomers to bitcoin think bitcoin core and bitcoin are different. The same could happen with dogecoin

Issue #1497, opened 2018-05-22
Doge wallet 1.10 Ubuntu 12,14 & 16 w/all lib's install error


ultra-pool:
`configure: error: libdb_cxx headers missing, Dogecoin Core requires this library for wallet functionality (--disable-wallet to disable wallet functionality)`

I have all the lib's installed. Tested on 2 VPS 1 full server and 4 VM's. Ubuntu 12,14 and 16, Mint and they all have the same error. All have different Lib releases. And all have the required lib's listed in the readme file and then some. And all tested on clean servers.

The small work around listed in the tickets here does not work. 

If you have also ran in this. Please reply. 

Thank you.

rllola:
I remember seeing this error in the past...
I have a fresh ubuntu 16 install. I will try to reproduce this.

I am mostly using docker when I want to use dogecoind (running 1.10). Maybe this can help you : 
https://github.com/BitcoinAmiens/docker-dogecoin

Will report tomorrow with more info.

rnicoll:
Err, which version of Berkeley DB do you have installed?

rllola:
I believe I have 5.3

rllola:
Nope my bad. I actually wasn't installed. But 5.1 doesn't seems available like in the documentation however I found 5.3 

So I did :
```
sudo apt install libboost-all-dev libdb5.3-dev libdb5.3++-dev
```

but now I have this error : 
```
checking for Berkeley DB C++ headers... default
configure: error: Found Berkeley DB other than 5.1, required for portable wallets (--with-incompatible-bdb to ignore or --disable-wallet to disable wallet functionality)
```
So I believe I have to download and build 5.1

rllola:
Actually no need to download 5.1 with incompatible db flag.

I have done a short gist. I hope it helps @ultra-pool 
https://gist.github.com/rllola/0e46feae5a41cb7d29352a84fb388304

@rnicoll If you think it is needed I can update the documentation ?

rnicoll:
I'm trying to decide if we want to update the docs, or move to 5.3/later. Probably yes can you raise a documentation update against the 1.14-branding branch, and we'll move to a later BDB for 1.15+.

rllola:
@rnicoll I believe this issue can be close now.

PR #1506, opened 2018-06-29
Update documentation for ubuntu and berkeley DB 5.3

rllola:
See https://github.com/dogecoin/dogecoin/issues/1497

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -80,7 +80,12 @@ install necessary parts of boost:
 
 BerkeleyDB is required for the wallet.
 
-    sudo apt-get install libdb5.3-dev libdb5.3++-dev libevent-dev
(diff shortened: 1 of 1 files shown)


rnicoll:
So... I haven't tried latest Ubuntu yet (oops, bad me), is 5.1 available for it? Normally we'd not recommend a version bump, but if it's the most sensible option we can of course do so.

Thanks for the PR!

rllola:
Berkeley 5.1 package is still available on Ubuntu 14.04 and I believe before. So I have re-added the old command and just add this for the Ubuntu 16.04 and later. If people prefer follow requirement and use 5.1 they can build it like describe in the berkeley section.

It start to have lots of notes...

rnicoll:
Looks good, thanks! Definitely thinking we'll move outright to 5.3 in 1.16.

PR #1523, opened 2018-09-19
Update miner confirmation window

rnicoll:
* Update miner confirmation window to 240 for Dogecoin main net, and update related values.
* Re-enable PoW tests with Dogecoin values.
* Update version bit tests based on Dogecoin mining values.


src/chainparams.cpp
```diff
@@ -238,8 +238,8 @@ class CTestNetParams : public CChainParams {
         consensus.nPowTargetTimespan = 4 * 60 * 60; // pre-digishield: 4 hours
         consensus.nPowTargetSpacing = 60; // 1 minute
         consensus.fPowNoRetargeting = false;
-        consensus.nRuleChangeActivationThreshold = 1512; // 75% for testchains
-        consensus.nMinerConfirmationWindow = 240; // nPowTargetTimespan / nPowTargetSpacing
+        consensus.nRuleChangeActivationThreshold = 7560; // 75% for testchains
+        consensus.nMinerConfirmationWindow = 10080; // 60 * 24 * 7 = 10,080 blocks, or one week
```
langerhans: I still think this is way too high for testnet. That means if we ever want to test bip9 deployments on testnet we need to mine it for over 5 days straight. Why not make this a day or two?

PR #1526, opened 2018-09-20
Fix CVE-2018-17144

Varunram:
Quoting bitcoincore.org, Security issue CVE-2018-17144: it was discovered that older versions of Bitcoin Core will crash if they try to process a block containing a transaction that attempts to spend the same input twice. Such blocks are invalid, so they can only be created by a miner willing to sacrifice their allowed income for creating a block of at least 12.5 BTC (about $80,000 USD as of this writing).

Since dogecoin core is a fork of bitcoin core, it is liable to the same bug. All dogecoin core versions equivalent to bitcoin core 0.14.0 and above need to upgrade in order to be immune to this vulnerability.

P.S: I don't know which branch to PR to, if this is the wrong branch, please update the branch.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2902,7 +2902,7 @@ bool CheckBlock(const CBlock& block, CValidationState& state, bool fCheckPOW, bo
 
     // Check transactions
     for (const auto& tx : block.vtx)
-        if (!CheckTransaction(*tx, state, false))
+        if (!CheckTransaction(*tx, state, true))
             return state.Invalid(false, state.GetRejectCode(), state.GetRejectReason(),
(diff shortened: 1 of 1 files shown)


Varunram:
Oops, wrong branch, the lead maintainer of bitcoin core merged it into the appropriate branch at https://github.com/dogecoin/dogecoin/commit/696b936aa3ab6f459d0e16f9805eaeb747a0036c.

langerhans:
Thanks for the work anyway! We are aware of the issue and Ross has already done the work to rebase our current 1.14 branch onto the latest Bitcoin Core 0.14 which contains this fix. That's why you found it there :) We have a plan for releasing 1.14 soon now so the fix should be out there in not too long.

PR #1544, opened 2018-10-17
Changes for mac installer branding - See notes

michilumin:
These changes are necessary for the deploy scripts to see the app correctly. Also includes our background image. Note that DS_Store should be generated, but I'm including it so that it should be there in any case; though it may need to be edited


rnicoll:
LGTM, but probably should have someone with a Mac eyeball it too. @langerhans  ?

langerhans:
bildschirmfoto 2018-11-04 um 15 58 12

Looks good (ignore the unsigned binary)

PR #1546, opened 2018-10-19
1.17 AuxPoW support

rnicoll:
Add AuxPoW support to 1.17.

This DOES NOT pass lint tests included in 1.17 Travis, and I'm not entirely sure what to do with that as I'd missed it early on, and it's going to take major changes to fix. Leaning on fixing 1.18, but wanted to flag in case you want to tell me it needs correcting before you'll accept.


rnicoll:
@langerhans Can you take a look at this please?

PR #1557, opened 2018-12-04
Update dogecoin.cpp

MuchVelociraptor:
Clearly this was what was meant in this comment.


langerhans:
Nice meme, but better keep the other version ;)

PR #1562, opened 2019-01-01
Update >>moon license year 2019

EvertonMelo:
### Fix license of use, 
update 2019; 
Format ISO 8869-1;
----------------------------------------------------------------------------------------------------------------------
in Memoriam Guto Schiavon 
guto-schiavon_bitcoin
R.I.P pioneer and friend. 

#ripGUTO

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,4 +1,4 @@
-Copyright (c) 2009-2015 The Bitcoin Core developers
+Copyright (c) 2009-2019 The Bitcoin Core developers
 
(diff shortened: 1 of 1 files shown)


langerhans:
Uhm, sorry for your loss, but this isn't the right branch, and generally we don't do trivial changes like that anyway ;) But thanks for the contribution.

EvertonMelo:
@langerhans 
ok. 

It was not just me that lost, the world lost. 

just as you missed the opportunity to continue in the past.

PR #1569, opened 2019-02-20
Create Dockerfile

MuchVelociraptor:
This doesn't work.


langerhans:
Why does it compile bdb manually? The dependency builder will already do that.

just-an-dev:
we can also use multistage image to have a very little image at end

PR #1572, opened 2019-03-09
Drop use of smart fees

rnicoll:
Drop use of smart fees when calculating transaction fees. Dogecoin fees are intentionally on a fixed schedule.


langerhans:
So this looks good, but should we actually at this point drop the -dev branch and put everything on -branding given that is quite a bit ahead already?

rnicoll:
I'll rebase branding on dev in a moment

PR #1587, opened 2019-06-09
Merge 1.14 into master

langerhans:
Up for debate is if we want to actually push almost 5k commits to master or somehow (don't ask me how) squash this. I'm not too keen on squashing everything as that would mean losing the "work log" of what has been done compared to upstream. On the plus side, this doesn't kill everyone's local master branch with a force push. I'm not against debating the force push tho.

This PR has one additional commit to update the issue template that would have otherwise had info about upstream.


rnicoll:
Anything I can do to help?

langerhans:
I was waiting for a good time to actually merge this cause we should do Gitian builds shortly after. Should be able to do it next week.

PR #1588, opened 2019-06-09
Merge master back into 1.10-maint

langerhans:
I know this is kinda useless given it's a commit and a revert, but for the sake of completeness...


langerhans:
Lol okay I didn't even notice, so nvm...

PR #1591, opened 2019-06-25
Added coments on chain parameters for regtest

rllola:
Make it a little bit more easy to read for people who are looking for those constants for development.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -139,7 +139,7 @@ public:
         digishieldConsensus.pLeft = &consensus;
         digishieldConsensus.pRight = &auxpowConsensus;
 
-        /** 
+        /**
          * The message start string is designed to be unlikely to occur in normal data.
          * The characters are rarely used upper ASCII, not valid as UTF-8, and produce
          * a large 32-bit integer with any alignment.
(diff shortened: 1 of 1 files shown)


src/chainparams.cpp
```diff
@@ -45,7 +45,7 @@ static CBlock CreateGenesisBlock(const char* pszTimestamp, const CScript& genesi
  * CBlock(hash=000000000019d6, ver=1, hashPrevBlock=00000000000000, hashMerkleRoot=4a5e1e, nTime=1386325540, nBits=0x1e0ffff0, nNonce=99943, vtx=1)
  *   CTransaction(hash=4a5e1e, ver=1, vin.size=1, vout.size=1, nLockTime=0)
  *     CTxIn(COutPoint(000000, -1), coinbase 04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73)
- *     CTxOut(nValue=50.00000000, scriptPubKey=0x5F1DF16B2B704C8A578D0B)
+ *     CTxOut(nValue=50.00000000, scriptPubKey=0x5F1DF16B2dogecoinB704C8A578D0B)
```
rnicoll: Err... is that even vaguely valid? Surely it has to be hex still?

rllola: Ah no my bad this is a bad manipulation that got me pasting the word 'dogecoin' in the code. This should not be here.... :sweat_smile:

rllola:
A more effective way to give this information would be in a doc or a wiki ? Any place on the world wide web where we could give this information ? So far I can find for Bitcoin easily but always have to look in the code or Dogecoin.

rnicoll:
There's a Dogecoin wiki on reddit, perhaps there? https://www.reddit.com/r/dogecoin/wiki/index

rnicoll:
I've rebased this against the `1.14-branding` branch we'd normally apply changes to, as https://github.com/dogecoin/dogecoin/pull/1599 - presume that gets merged I'll then close both.

rllola:
The wiki on reddit is not really technical but as some useful links. I could certainly write a gist and linked it there. But do you think it could be more practical to have a wiki on the github ?

rllola:
Ok I just find out that there is already a wiki but also that everyone can modify it ???

I started a page : https://github.com/dogecoin/dogecoin/wiki/Dogecoin-special-parameters

rnicoll:
Hmmm... edit controls on Github wikis are a nightmare, they're either public or people with push permission on the repo, nothing finer grain it seems. That's not ideal. Let me ponder

langerhans:
I have manually merged the commits into 1.14-maint. Thx!

PR #1596, opened 2019-07-12
Update to litecoin_scrypt

rnicoll:
The old `ltc_scrypt` module works on systems it is installed on, but does not install on new systems it appears. Our README says to use `litecoin_scrypt`, so updating the code to match.


langerhans:
Manually merged in https://github.com/dogecoin/dogecoin/commit/5d58c691caf4fb9fe829ddaa54b213bc81ac5f33

PR #1600, opened 2019-07-13
Revert behaviour change in getauxblock

rnicoll:
Change getauxblock to respond with a "target" field instead of "_target".


langerhans:
Manually merged into 1.14-maint

PR #1602, opened 2019-07-14
Update 1.17 to the point where it syncs fully

rnicoll:


rnicoll:
Abandoning in preference for a more recent version

PR #1706, opened 2021-02-06
fix minimum required space for data directory

slightlyskepticalpotat:
The minimum required space for the data directory (shown when first launching the qt wallet) was originally unchanged from Bitcoin. This reduces it to just above the current size of the Dogecoin blockchain.

### DIFF
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -23,7 +23,7 @@
 
 static const uint64_t GB_BYTES = 1000000000LL;
 /* Minimum free space (in GB) needed for data directory */
-static const uint64_t BLOCK_CHAIN_SIZE = 50;
+static const uint64_t BLOCK_CHAIN_SIZE = 55;
 /* Minimum free space (in GB) needed for data directory when pruned; Does not include prune target */
 static const uint64_t CHAIN_STATE_SIZE = 2;
 /* Total required space (in GB) depending on user choice (prune, not prune) */
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -23,7 +23,7 @@
 
 static const uint64_t GB_BYTES = 1000000000LL;
 /* Minimum free space (in GB) needed for data directory */
-static const uint64_t BLOCK_CHAIN_SIZE = 120;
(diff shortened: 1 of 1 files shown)


DemonRx:
Is this intro statement even necessary? Being that the standard storage medium in the average home PC is 500gb+. It'll be probably more than a decade before that might be even reached.

Or could this be made dynamic so it updates the statement to the user automatically?  As it seems redundant having to update this value constantly.

DemonRx:
@slightlyskepticalpotat the commit name says 'maximum' vs your PR says 'minimum' probably want to fix that typo.

slightlyskepticalpotat:
Thanks for catching that, I've fixed it. I think the intro statement is still needed now, because having a core wallet taking up 1/10 of your hard drive is still a lot, especially if you could use a lighter wallet. Additionally, downloading 50 GB of data still takes a while.

rnicoll:
Verified disk usage on one of my nodes is ~46GB so this gives a working overhead for now.

rnicoll (approved):

shibe2:
With txindex it occupies about 52 GB on my disk. Note that decimal prefix is used here.

rnicoll:
> Or could this be made dynamic so it updates the statement to the user automatically? As it seems redundant having to update this value constantly.

In terms of updates, we have to review patches as we update each release anyway, so we'll review then. Although I'm _hoping_ a later Bitcoin Core makes this depend on block max size & checkpoints or something.

src/qt/intro.cpp
```diff
@@ -23,7 +23,7 @@
 
 static const uint64_t GB_BYTES = 1000000000LL;
 /* Minimum free space (in GB) needed for data directory */
-static const uint64_t BLOCK_CHAIN_SIZE = 120;
+static const uint64_t BLOCK_CHAIN_SIZE = 50;
```
rnicoll, line 26:
```suggestion
static const uint64_t BLOCK_CHAIN_SIZE = 55;
```
Lets have a bit more space just to be sure

slightlyskepticalpotat: It uses 47 GB on my node, but I don't have txindex.

patricklodder (approved):
I think this is good for this release.

PR #1745, opened 2021-02-17
Bump version to 1.14.3

rnicoll:


patricklodder (approved):

doc/release-notes.md
```diff
@@ -1,6 +1,6 @@
 Dogecoin Core version 1.14.2 is now available from:
 
-  <https://github.com/dogecoin/dogecoin/releases/tag/v1.14.0/>
+  <https://github.com/dogecoin/dogecoin/releases/tag/v1.14.2/>
```
IncognitoJam, line 3: Should these be 1.14.3?

rnicoll: It's been a bit of a discussion, but I believe the conclusion is we don't point to a dead link, yet. We will when preparing final relaese notes.

PR #1761, opened 2021-02-24
Target  OS non-windows use fPIC

DemonRx:
Revisiting this as it was previously a reverted merge because it wasn't done to the dev branch. (see, #1420, #1447)

## Scope of PR:

Resolves an issue where the compiled binaries (in particular dogecoin-qt) cannot be executed using the mouse in non windows environments (i.e. Linux). Forcing users to have to start the client from terminal.

By building with fPIE, the resulting objects are "LSB shared objects".  Thus changing to fPIC builds a LSB executable binary that a user can double-click on to start the program.

This PR changes `-fPIE to fPIC`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -514,8 +514,8 @@ if test x$use_hardening != xno; then
   AX_CHECK_LINK_FLAG([[-Wl,-z,now]], [HARDENED_LDFLAGS="$HARDENED_LDFLAGS -Wl,-z,now"])
 
   if test x$TARGET_OS != xwindows; then
-    AX_CHECK_COMPILE_FLAG([-fPIE],[PIE_FLAGS="-fPIE"])
-    AX_CHECK_LINK_FLAG([[-pie]], [HARDENED_LDFLAGS="$HARDENED_LDFLAGS -pie"])
+    AX_CHECK_COMPILE_FLAG([-fPIC],[PIE_FLAGS="-fPIE"])
+    AX_CHECK_LINK_FLAG([[-fno-pie]], [HARDENED_LDFLAGS="$HARDENED_LDFLAGS -pie"])
   fi
 
   case $host in


DemonRx:
ping

patricklodder:
There was a reason for keeping fPIE on non-windows builds, iirc something with the deterministic build process, but it looks like on later bitcoin versions it's only enforced for OSX, with this particular statement completely removed. I'm trying to figure out what exactly changed upstream and why, and if we can fix this without breaking the deterministic builds that we need for gitian.

patricklodder:
Update: I'm having a hard time making sense of the upstream changes to this, it seems all over the place. We'll have to skip this for 1.14.3, as I cannot get this fully done on the critical path, without risking a broken gitian build.

However, we have a set of improvements to do post 1.14.3 with other things we could not get done in time, so I'll make a board for that today and target resolution of this to be included in that effort.

kitcateleven:
hellopeeps!

DemonRx:
Posting this for reference: https://github.com/bitcoin/bitcoin/issues/3847 ( build as "ELF 64-bit LSB shared object" instead of "ELF 64-bit LSB executable" )


Compiling using PIE is for the purpose of hardening the binary as explained here:

>Another difference is that executables have a defined entry point address offset, i.e., 0x08048000 for i386, 0x00400000 for x86 and 0x00010000 for arm.
>
>A shared object file can be a library, but also an executable. When being an executable, there is no such offset. A shared object executable, so to say, is a positional independent executable (PIE) using address space layout randomization (ASLR). Thus, when looking at its /proc/pid/maps file, you will notice that the location of the loaded segments vary in each execution in contrast to standard executables.
>
>The idea behind this feature is to add security to executables by hindering attackers from performing return-oriented programmings attacks. Many maintainers decided to build packages with PIE enabled as the default, e.g., since Fedora 23 or with Ubuntu 17.10.

- https://askubuntu.com/a/960212

I was not aware of this. Learned something new today. 

However, what is interesting, by default in Ubuntu 16, ASLR is not enabled by default in `kernel.randomize_va_space = 1`, but in Ubuntu 18 it is (kernel.randomize_va_space = 2). (I'm not sure about 17 or 20). 

In any case it seems the better choice is to keep this compiling option as-is, and explore if it's possible to make it more mouse friendly (as a matter of convenience).

patricklodder:
There is a `.desktop` file in [contrib/debian](https://github.com/dogecoin/dogecoin/tree/038ccec3e44dbfa458e00aff99a8a27fac7fb8d7/contrib/debian) that should improve "clickability"

DemonRx:
@patricklodder nice, worked perfectly. 

In installing the desktop icon. I do want to point out that it references the pixmap icon file as "dogecoin128.png" https://github.com/dogecoin/dogecoin/blob/038ccec3e44dbfa458e00aff99a8a27fac7fb8d7/contrib/debian/dogecoin-qt.desktop#L11 However, in the src tree the file is named [bitcoin128.png](https://github.com/dogecoin/dogecoin/blob/038ccec3e44dbfa458e00aff99a8a27fac7fb8d7/share/pixmaps/bitcoin128.png) . So I had to create a symlink with dogecoin's name to the bitcoin png.  Ofc I could change the naming in the .desktop file too. But unless this is changed in the repo, the link is more sensible for the time being. 

- Should we rename the pixmap file  ? As this also causes an interference with installing bitcoin-core that actually uses bitcoin128.png

I also came across that installing the .desktop file it is using an encoding key which is deprecated. And can be safely removed per https://lintian.debian.org/tags/desktop-entry-contains-encoding-key.html
https://github.com/dogecoin/dogecoin/blob/038ccec3e44dbfa458e00aff99a8a27fac7fb8d7/contrib/debian/dogecoin-qt.desktop#L2

I validated this by running the command:

`desktop-file-validate /usr/share/applications/dogecoin-qt.desktop`

It's output states:
```bash
/usr/share/applications/dogecoin-qt.desktop: warning: key "Encoding" in group "Desktop Entry" is deprecated
```
lmk your thoughts. I can start a PR one or both of those. Or if you want to take care of it, or just leave it.

patricklodder:
> Should we rename the pixmap file ? As this also causes an interference with installing bitcoin-core that actually uses bitcoin128.png

From where I'm sitting an image depicting the Dogecoin logo should **_not_** be referred to as "bitcoin", so I'd change the .png filenames. @rnicoll / @langerhans any objection?

> key "Encoding" in group "Desktop Entry" is deprecated

Since UTF-8 is a requirement, sounds like a great plan to remove that line :-)

kitcateleven:
As long as you think it makes the system work better, go for it.  Thanks so much for all of your hard work

Thank you!

Cat Mayville, President
Spot Digital Media
404-357-4503

DemonRx:
I'll close this PR in favor of the solution indicated by @patricklodder https://github.com/dogecoin/dogecoin/pull/1761#issuecomment-787577010

I'll start a PR on the deprecated encoding key. In regards to the PNG images. I think there might be other references to them in other areas of the code. So that might not be a simple change of just renaming the files. I'll wait @rnicoll and @langerhans input before pursuing.

PR #1765, opened 2021-02-26
Formatting in "Development Tips and Tricks"

leuqarte:
Formatting the `commands` within this section


leuqarte:
submitting against `1.14.3-dev` instead

PR #1800, opened 2021-03-15
Optimize Images With ImgBot

slightlyskepticalpotat:
Achieves a 20.57% reduction in total image size with lossless compression.


DemonRx:
https://github.com/dogecoin/dogecoin/tree/master/share/pixmaps these as well ?

slightlyskepticalpotat:
The bot I used leaves images that can't be optimized without quality loss alone, so those didn't leave any room for optimization. As a side note, it also optimized the PNG images. I'm not sure why the git diff shows them as empty files.

patricklodder (commented):
Looks good to me. Normally I'd ask to split doc from code commits on this, but since we're on the tail end of a release cycle, I really don't care too much. Leaving it up to @rnicoll

rnicoll:
@slightlyskepticalpotat For the images which aren't shipped (everything outside of `src/qt/res/` this is actually likely to make things worse, as I believe Git will download the full history, not just the latest image. What I mean is, everyone grabbing the source code will end up with the previous large image, and it's smaller replacement.

Could you split out just the `src/qt/res/` changes, as they're easier to approve, at least? We can then address the contrib and docs changes later.

slightlyskepticalpotat:
> @slightlyskepticalpotat For the images which aren't shipped (everything outside of `src/qt/res/` this is actually likely to make things worse, as I believe Git will download the full history, not just the latest image. What I mean is, everyone grabbing the source code will end up with the previous large image, and it's smaller replacement.
> 
> Could you split out just the `src/qt/res/` changes, as they're easier to approve, at least? We can then address the contrib and docs changes later.

Done, now only the src/qt/res/ changes are included in this PR.

rnicoll (approved):

patricklodder (approved):

rnicoll:
Thank you!

PR #1825, opened 2021-04-16
[QT] Add a form dialog for adding peers

alamshafil:
This is my first time contributing to a open source project! I implemented the form dialog to add peers. I am fairly new to Qt and C++, so the code may not be well optimized. 

dogecoin_1
dogecoin_2

### DIFF
--- a/src/qt/peerdialog.cpp
+++ b/src/qt/peerdialog.cpp
@@ -1,0 +1,176 @@
+// Copyright (c) 2011-2016 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "guiutil.h"
+#if defined(HAVE_CONFIG_H)
+#include "config/bitcoin-config.h"
+#endif
+
+#include <iostream>
+#include <string>
+
+#include "peerdialog.h"
+
+#include "ui_addpeerdialog.h"
+#include "ui_testpeerdialog.h"
+
+#include "net.h"
+#include "net_processing.h"
+#include "netbase.h"
+#include "protocol.h"
+#include "chainparams.h"
+#include "util.h"
+
+#include <stdio.h>
+
+#include <QMessageBox>
+#include <QHostAddress>
+#include <QAbstractSocket>
+#include <QUrl>
+
+/** Function to manage peers */
+QString PeerTools::ManagePeer(QString type, QString peer)
+{
+    std::string peerAddress = peer.toStdString();
+
+    if(!g_connman)
+        return "Error: Peer-to-peer functionality missing or disabled";
+
+    if (type == "onetry")
+    {
+        CAddress addr;
+        g_connman->OpenNetworkConnection(addr, false, NULL, peerAddress.c_str());
+        return "Attempted to one try node.";
+    }
+
+    if (type == "add")
+    {
+        if(!g_connman->AddNode(peerAddress))
+            return "Error: Node already added";
+    }
+    else if(type == "remove")
+    {
+        if(!g_connman->RemoveAddedNode(peerAddress))
+        {
+            if(!g_connman->DisconnectNode(peerAddress))
+                return "Node not found in connected nodes";
+
+            return "Disconnected the node: " + peer;
+        }
+        else
+        {
+            if(!g_connman->DisconnectNode(peerAddress))
+                return "Node not found in connected nodes";
+        }
+    }
+
+    return "Returned OK.";
+}
+
+/** Check if Peer is valid */
+bool PeerTools::CheckPeerAddress(QString address)
+{
+    CNetAddr addr;
+    return LookupHost(address.toStdString().c_str(), addr, true);
+}
+
+/** Get port based on current chain */
+QString PeerTools::GetPort()
+{
+    return QString::number(Params().GetDefaultPort());
+}
+
+/** Add Peer Dialog */
+AddPeerDialog::AddPeerDialog(QWidget *parent) : 
+    QWidget(parent),
+    ui(new Ui::AddPeerDialog)
+{
+    ui->setupUi(this);
+
+    ui->peerPort->setValidator( new QIntValidator(1, 65535, this) );
+
+    connect(ui->pushButton, SIGNAL(clicked()), this, SLOT(on_addPeer_clicked()));
+}
+
+AddPeerDialog::~AddPeerDialog()
+{
+    delete ui;
+}
+
+void AddPeerDialog::on_addPeer_clicked()
+{
+    QString address = ui->peerAddress->text();
+    QString port = ui->peerPort->text();
+    QString data = "";
+
+    if(address.isEmpty()) 
+    {
+        QMessageBox::critical(this, "Add Peer", "Please enter an address.", QMessageBox::Ok, QMessageBox::Ok);
+        return;
+    }
+
+    if(port.isEmpty()) 
+    {
+        port = PeerTools::GetPort();
+        ui->peerPort->setText(port);
+    }
+
+    if(!PeerTools::CheckPeerAddress(address))
+    {
+        QMessageBox::critical(this, "Add Peer", "Please enter a vaild peer address.", QMessageBox::Ok, QMessageBox::Ok);
+        return;
+    }
+
+    data = address + ":" + port;
+
+    if(QMessageBox::Ok == QMessageBox::information(this, "Add Peer", PeerTools::ManagePeer("add", data), QMessageBox::Ok, QMessageBox::Ok))
+        this->close();
+}
+
+/** Add Test Peer Dialog */
+TestPeerDialog::TestPeerDialog(QWidget *parent) : 
+    QWidget(parent),
+    ui(new Ui::TestPeerDialog)
+{
+    ui->setupUi(this);
+
+    ui->peerPort->setValidator( new QIntValidator(1, 65535, this) );
+
+    connect(ui->pushButton, SIGNAL(clicked()), this, SLOT(on_testPeer_clicked()));
+}
+
+TestPeerDialog::~TestPeerDialog()
+{
+    delete ui;
+}
+
+void TestPeerDialog::on_testPeer_clicked()
+{
+    QString address = ui->peerAddress->text();
+    QString port = ui->peerPort->text();
+    QString data = "";
+
+    if(address.isEmpty()) 
+    {
+        QMessageBox::critical(this, "Test Peer", "Please enter an address.", QMessageBox::Ok, QMessageBox::Ok);
+        return;
+    }
+
+    if(port.isEmpty()) 
+    {
+        port = PeerTools::GetPort();
+        ui->peerPort->setText(port);
+    }
+
+    if(!PeerTools::CheckPeerAddress(address))
+    {
+        QMessageBox::critical(this, "Test Peer", "Please enter a vaild peer address.", QMessageBox::Ok, QMessageBox::Ok);
+        return;
+    }
+
+    data = address + ":" + port;
+
+    if(QMessageBox::Ok == QMessageBox::information(this, "Try Peer", PeerTools::ManagePeer("onetry", data), QMessageBox::Ok, QMessageBox::Ok))
+        this->close();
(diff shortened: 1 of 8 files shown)


alamshafil:
I have two questions:
What is the default network node port?
How do you interact with the RPC to perform commands in Qt C++? As of now I just used the RPCExecutor from the RPC Console.

alamshafil:
Also, I accidentally messed the git up by force-pushing... 
So I forced it back to the commit 407dcd4, which was before I first did the first force push.
If this causes any problem please let me know.

patricklodder:
closing & reopening to trigger CI against 1.14.4-dev

alamshafil:
Confirmed it works on Linux x64 using version 1.14.4-dev.

src/qt/peerdialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+    QString peer = ui->lineEdit->text();
+    QString data = "";
+
+    if(peer.isEmpty()) 
+    {
+         QMessageBox::critical(this, "Add Peer", "Please enter an address.", QMessageBox::Ok, QMessageBox::Ok);
+         return;
+    }
+
+    data = peer.contains(":") ? peer : peer + ":" + "22556";
```
patricklodder, line 88: 2 issues with this:

1. an ipv6 peer has colons in the address part
2. standard port is 44556 for testnet

Perhaps it's user-friendlier to make 2 input fields - one for address, one for port - and remove the string interpretation altogether?

(same for line 120)

alamshafil: You are right about the ipv6 address. I will add the 2 input fields in next commit.

alamshafil:
I have added the two input fields for the peer address and peer port. I believe the user needs to be aware of the default network ports, so I have added it as a message if you leave the port field blank.

dogecoin_peer_1
dogecoin_peer_2

michilumin:
Hey so, taking a look at this while Pat is otherwise occupied. I think this is really good, and it's working rather well.  There are just a few things I'm wondering if you'd be willing to give a stab at before merge;

1) So we probably want to validate the IP addresses before executing the command. I believe the approach you'd want to take with this in Qt is pretty simple, putting your ip address into for example 'incomingString', and then checking for true or false on QAbstractSocket::IPv4Protocol == incomingString.protocol();  and QAbstractSocket::IPv6Protocol == incomingString.protocol();  If they == eachother, then continue; if not, then display an error and kick it back to the user.

2) for Remove Peer, it'd be super nice if there wasn't a dialog at all; but whatever NodeID was highlighted, it'd grab the IP from the Node/Service column and do a remove on that one.  It looks like this would be retrieved through GUIUtil::getEntryData(ui->peerWidget, PeerTableModel::Address); which you could then just feed to your removal code.

3) If you can, maybe just return to the user instead of just 'executed command', what the response of the rpc command is? If its just 'null', then you can check for that and respond that it returned OK. But if it's any other error, (anything but 'null') maybe make it an error box instead and just forward that error on to the user.


... I hope this isn't too much to ask; because these kinds of adds are really, really what the core wallet needs. If this seems to be too much, please let me know but - also let me know what you think?

shibe2:
I think, it would be good if the accepted address format was the same as for addnode, connect, seednode configuration parameters and addnode RPC. Then you could have a peer list that works for both UI and non-UI users. Consequently, validation should accept all address formats that would work for other ways of entering peer addresses. That includes IP literals, DNS names, onion addresses. If you want separate fields for host and port, you should also allow pasting combined address and port string as accepted in other places.

alamshafil:
Thank you so much for the feedback @michilumin! I have no problem working on the stated suggestions. Here is what I think about them:

1. I can validate the IP address no problem. It is important that a valid IP address is passed in.
2. I thought of doing this when I started this, but my limited QT skills didn't allow me to do it. Thank you for leading me in the right direction.
3. The code already tells the user if the node has been already added or any other error. I was not sure how to interpret the `null` response returned, so I put `Executed command.`

I will push commits for each of these issues soon!

alamshafil:
Thank you @shibe2 for the suggestions.

1. I believe I can add DNS names, IP addresses, and onion addresses. I am not sure how I would go validating onion links. Does the RPC console accept onion links?  
2. Also, how would the user paste a peer with the port? Would it go in the first input box? Do you think we should revert to the one input field and correctly determine the port?

alamshafil:
dogecoin_remove

michilumin:
Hey @alamshafil - Checked out this latest commit; and, much improved, very nice;  just two things I noticed:

1) IP format checking doesn't seem to be 'complete', and looking further into it, this is a known issue with .protocol().  as if i just enter '125' for the IP address it still returns OK.  (it would populate the address as 125.0.0.0.) - I think a decent way to pre-screen for this (Still do the .protocol() check!) would be to go through your string and look for either one ':' or four '.' (no more  no less if==),  and a length of at least 7.  If it finds a : pass it to the validator code that you have; and if it passes the len >= 7 check and number of '.' == 4 check,  if that passes feed that over to the .protocol() check. 

2) Remove, for me, seems to work as long as I added the node. Now, this is also known behavior in most QT wallets. For nodes that have been discovered, vs added with addnode, it always returns 'Error: Node has not been added'. Happens in the console too.  Though maybe if this occurs, (if that error returns) your system should attempt a 'disconnectnode' instead? Since it's very likely that folks using this interface who are removing a node, just want it disconnected immediately. 

Finally: Adding a peer with the port as @shibe2 mentions gets a little trickier, since you'd have to get pretty deep into parsing the string- its usually denoted as :123.245.123.245:PORT  so  you'd have to parse the four '.' delimiters and then the ':'.   

The capabilities to add by DNS name I think is important, but I think less important is 'exactly mirroring what the RPC console' takes in.  Address and port is pretty straightforward to the user. This kind of being an offshoot of making the UI friendlier I think there could be, later on maybe, another button 'Add Node List' that pops up a freeform text box and specifically strides the 'addnode ip.add.re.ss:port' format that appears on most block explorers and performs the addnodes on all of those.

shibe2:
It is possible to add onion node address via RPC. The syntax is a subset of DNS name.

If you make separate fields, combined host:port address should be pasteable into host field.

I think that invalid IP literals should be rejected everywhere: UI, RPC, configuration, command line.

If you agree with my argument about supporting existing address formats in the UI, then it is important to support that from the beginning to avoid creating incompatibilities. For example one can write:

>add my node 192.0.2.6:1234

If any version of Dogecoin Core doesn't support that format, one may have to write:

>add my node address 192.0.2.6 port 1234 from UI or 192.0.2.6:1234 in command line, configuration file, or RPC

alamshafil:
I used a regular expression to validate IP addresses because it seemed like easier way to validate them. I also used a regular expression to validate DNS names, such as `example.com` or `zqktlwi4fecvo6ri.onion`. It would be useful to rigoursly test these regular expressions to see if they work correctly when validating IP addresses or DNS names.

src/qt/peerdialog.cpp
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+
+    return "Returned OK.";
+}
+
+/** Check if Peer is vaild */
+bool PeerTools::CheckPeerAddress(QString address)
+{
+    if(PeerTools::CheckIPAddress(address))
```
rnicoll, line 74: I haven't tested this to be sure, but would `LookupHost()` in netbase.cpp be a good way to validate the address perhaps? It looks like it basically is how the node will resolve it anyway, so would be identicalish validation to what happens later.

src/qt/peerdialog.cpp
```diff
@@ -0,0 +1,233 @@
...
+    if (chain == "main") {
+        return "22556";
+    } else if (chain == "test") {
+        return "44556";
+    } else if (chain == "regtest") {
+       return "18332";
+    } else {
+        /** Chain type is unknown so return main chain port */
+        return "22556";
+    }
```
rnicoll, line 138:
```suggestion
    return QString::number(Params().GetDefaultPort());
```

src/qt/peerdialog.cpp
```diff
@@ -0,0 +1,233 @@
...
+        {
+            if(!g_connman->DisconnectNode(peerAddress))
+                return "Node not found in connected nodes";
+        }
+    }
+
+    return "Returned OK.";
+}
+
+/** Check if Peer is vaild */
```
rnicoll, line 71:
```suggestion
/** Check if Peer is valid */
```

src/qt/peerdialog.cpp
```diff
@@ -0,0 +1,233 @@
...
+        else
+            return false;
+    }
+    else if (QAbstractSocket::IPv6Protocol == address.protocol())
+    {
+        return true;
+    }
+    else
+    {
+        return false;
```
rnicoll, line 120:
```suggestion
    CNetAddr addr;
    return LookupHost(address.toStdString().c_str(), addr, true);
```

Can you try this, it seems to work for me, but I'm not sure I haven't missed a corner case somewhere.

alamshafil:
Thank you @rnicoll for the suggestions! There was a typo which included a extra `}` so I fixed it for it to compile. I have committed them to the pull request. The `LookupHost()` code works, but it accepts numbers like `127.0`. It could be a problem with converting the `QString` to `std::string`. I could be wrong about that.

alamshafil:
I had the same issue when using `QHostAddress` to validate the address, it would return true if it was `127`. Therefore I used two methods of verification. (QHostAddress and a regex).

rnicoll:
> 
> 
> I had the same issue when using `QHostAddress` to validate the address, it would return true if it was `127`. Therefore I used two methods of verification. (QHostAddress and a regex).

@michilumin I'd appreciate your thoughts here, I'm a bit "Hey if the resolver thinks it's valid, then I'm okay with it", but maybe I shouldn't be. I'm definitely struggling to think if clear criteria which are both open enough for every scenario and still useful.

alamshafil:
I believe we should focus on user-experience when deciding the final outcome. I also believe if the code thinks the address is valid, then it's 100% valid. But some novice user might get confused how their random number got accepted. I'm struggling to see the best option here.

michilumin:
So here's the thing:  UX is king, *however* we can't guide entry completely.  While I don't really believe that entering, say, '127' is equivalent to 127.0.0.0;  the libs we're using for this provide that shortcut functionality.  I think combining both regex and Qt's validation are probably far enough to go;  entering "127" and getting it accepted basically is out of the scope of our UX guidance, because it's completing a wrong entry.  Format wise, it isn't incorrect, as it's shorthand for 127.0.0.0 according to the method we're using to capture user input.   In short, if a user enters '127' and it gets accepted without being able to realize the error, the problem at play is beyond our validation methods.  So I think this is OK; as it still prevents entry of a completely malformed non-feasible address.

sabotagebeats:
> I think, it would be good if the accepted address format was the same as for addnode, connect, seednode configuration parameters and addnode RPC. Then you could have a peer list that works for both UI and non-UI users. Consequently, validation should accept all address formats that would work for other ways of entering peer addresses. That includes IP literals, DNS names, onion addresses. If you want separate fields for host and port, you should also allow pasting combined address and port string as accepted in other places.

is there currently compatibility for loading a list of peers? I find it is very cumbersome to add nodes one at a time, especially when needing to paste multiple fields per node.

rnicoll (approved):

patricklodder:
Is it possible to rebase and squash these commits please?

See [squashing in our shiny new contribution guide](https://github.com/dogecoin/dogecoin/blob/1.14.4-dev/CONTRIBUTING.md#squashing-commits)

patricklodder:
> is there currently compatibility for loading a list of peers? I find it is very cumbersome to add nodes one at a time, especially when needing to paste multiple fields per node.

There are only 8 addnode slots, but if you have an example of what you're doing, could you please open a new issue for discussion?

alamshafil:
@patricklodder I have squashed the commits.

patricklodder (approved):
utACK

rnicoll:
Putting this on my re-test queue for tonight.

rnicoll (approved):
ACK

rnicoll:
Thanks!

alamshafil:
Glad I could help out!

PR #1834, opened 2021-04-18
Travis: Use minimal trusty image

langerhans:
This removes a ton of unneeded external apt repos out of which at least one (cassandra) is broken right now.


patricklodder (approved):
A minimal change, saying minimal, from the King of minimalism. ACK ❤️

PR #1856, opened 2021-04-25
Qt Add GUI wif import dialog

chromatic:
Addresses #1808.

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -102,6 +102,7 @@ BitcoinGUI::BitcoinGUI(const PlatformStyle *_platformStyle, const NetworkStyle *
     sendCoinsMenuAction(0),
     usedSendingAddressesAction(0),
     usedReceivingAddressesAction(0),
+    importPrivateKeyAction(0),
     signMessageAction(0),
     verifyMessageAction(0),
     aboutAction(0),
@@ -392,6 +393,9 @@ void BitcoinGUI::createActions()
     openAction = new QAction(platformStyle->TextColorIcon(":/icons/open"), tr("Open &URI..."), this);
     openAction->setStatusTip(tr("Open a dogecoin: URI or payment request"));
 
+    importPrivateKeyAction = new QAction(platformStyle->TextColorIcon(":/icons/address-book"), tr("&Import Private Key..."), this);
+    importPrivateKeyAction->setStatusTip(tr("Import a Dogecoin private key"));
+
     showHelpMessageAction = new QAction(platformStyle->TextColorIcon(":/icons/info"), tr("&Command-line options"), this);
     showHelpMessageAction->setMenuRole(QAction::NoRole);
     showHelpMessageAction->setStatusTip(tr("Show the %1 help message to get a list with possible Dogecoin command-line options").arg(tr(PACKAGE_NAME)));
@@ -418,6 +422,7 @@ void BitcoinGUI::createActions()
         connect(usedReceivingAddressesAction, SIGNAL(triggered()), walletFrame, SLOT(usedReceivingAddresses()));
         connect(openAction, SIGNAL(triggered()), this, SLOT(openClicked()));
         connect(paperWalletAction, SIGNAL(triggered()), walletFrame, SLOT(printPaperWallets()));
+        connect(importPrivateKeyAction, SIGNAL(triggered()), walletFrame, SLOT(importPrivateKey()));
     }
 #endif // ENABLE_WALLET
 
@@ -445,6 +450,7 @@ void BitcoinGUI::createMenuBar()
         file->addAction(verifyMessageAction);
         file->addAction(paperWalletAction);
         file->addSeparator();
+        file->addAction(importPrivateKeyAction);
         file->addAction(usedSendingAddressesAction);
         file->addAction(usedReceivingAddressesAction);
         file->addSeparator();
--- a/src/qt/forms/importkeysdialog.ui
+++ b/src/qt/forms/importkeysdialog.ui
@@ -1,0 +1,208 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>ImportKeysDialog</class>
+ <widget class="QDialog" name="ImportKeysDialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>560</width>
+    <height>200</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Import Private Key</string>
+  </property>
+  <property name="modal">
+   <bool>true</bool>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <widget class="QWidget" name="tabImportPrivateKey">
+     <attribute name="title">
+      <string>&amp;Import Private Key</string>
+     </attribute>
+     <layout class="QVBoxLayout" name="verticalLayout_Display">
+      <item>
+       <layout class="QHBoxLayout" name="horizontalLayout_1_ImportPrivateKey">
+        <item>
+         <widget class="QLabel" name="privateKeyWidgetLabel">
+          <property name="text">
+           <string>Private Key:</string>
+          </property>
+          <property name="textFormat">
+           <enum>Qt::PlainText</enum>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QLineEdit" name="privateKey">
+          <property name="toolTip">
+           <string>Private key to import into your wallet</string>
+          </property>
+         </widget>
+        </item>
+       </layout>
+      </item>
+      <item>
+       <layout class="QHBoxLayout" name="horizontalLayout_2_ImportPrivateKey">
+        <item>
+         <widget class="QLabel" name="privateKeyLabelWidgetLabel">
+          <property name="toolTip">
+           <string>Label for this private key in your wallet</string>
+          </property>
+          <property name="text">
+           <string>Label (optional):</string>
+          </property>
+          <property name="buddy">
+           <cstring>privateKeyLabel</cstring>
+          </property>
+         </widget>
+        </item>
+        <item>
+         <widget class="QLineEdit" name="privateKeyLabel">
+          <property name="toolTip">
+           <string>Label for this private key in your wallet</string>
+          </property>
+         </widget>
+        </item>
+       </layout>
+      </item>
+      <item>
+       <widget class="QCheckBox" name="rescanCheckBox">
+        <property name="toolTip">
+         <string extracomment="Rescan the blockchain on import (disabled when running in pruned mode)."/>
+        </property>
+        <property name="text">
+         <string>Rescan</string>
+        </property>
+       </widget>
+      </item>
+      <item>
+       <widget class="QLabel" name="privateKeyImportTextMessage">
+        <property name="text">
+         <string></string>
+        </property>
+       </widget>
+      </item>
+      <item>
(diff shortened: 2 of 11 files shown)


src/qt/bitcoingui.cpp
```diff
@@ -392,6 +393,9 @@ void BitcoinGUI::createActions()
     openAction = new QAction(platformStyle->TextColorIcon(":/icons/open"), tr("Open &URI..."), this);
     openAction->setStatusTip(tr("Open a dogecoin: URI or payment request"));
 
+    importPrivateKeyAction = new QAction(platformStyle->TextColorIcon(":/icons/address-book"), tr("&Import Private Key..."), this);
+    importPrivateKeyAction->setStatusTip(tr("Import a dogecoin private key"));
```
rnicoll, line 397:
```suggestion
    importPrivateKeyAction->setStatusTip(tr("Import a Dogecoin private key"));
```

src/qt/forms/importkeysdialog.ui
```diff
@@ -0,0 +1,226 @@
...
+           </property>
+           <property name="textFormat">
+            <enum>Qt::PlainText</enum>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QLineEdit" name="privateKey">
+           <property name="toolTip">
+            <string>Private Key to import into your wallet</string>
```
rnicoll, line 45:
```suggestion
            <string>Private key to import into your wallet</string>
```

rnicoll (changes requested):
Couple of very minor nit-picks, and I need to actually try running it!

src/qt/forms/importkeysdialog.ui
chromatic: Addressed this too.

rnicoll:
Main issue is this does a rescan on the GUI thread, so the whole interface hangs. For a rescan can you perhaps open a dialog that shows the rescan is in progress, but without freezing the whole UI? Could be a modal dialogue to protect against anyone triggering multiple rescans.

chromatic:
> Main issue is this does a rescan on the GUI thread, so the whole interface hangs. For a rescan can you perhaps open a dialog that shows the rescan is in progress, but without freezing the whole UI? Could be a modal dialogue to protect against anyone triggering multiple rescans.

I'll look into that. The current behavior duplicates the console window behavior, so it may be worth extracting the key behavior into key code shared between these two import mechanisms. I don't know if that should or shouldn't be in scope of this PR though.

rnicoll:
My inclination is get it working well, then clean up behind the scenes.

Let me know if you would like a hand with doing the rescan on a different thread, I haven't done Qt work in ages, but happy to take a look at least.

chromatic:
Thanks! I'll take the approach you suggest.

I'll look into the thread work and follow up with you if I get stuck. If you or anyone else reading this can quickly point to code that does something similar, that might be enough to get me going in the right direction (but I wouldn't spend time digging for it; if it's there, I'll find it).

chromatic:
@rnicoll I've just pushed code that accomplished this, though it's a little less elegant than I'd like. The happy paths seem to work okay in my testing.

Germanstreetstyle (approved):

Germanstreetstyle (approved):

Germanstreetstyle (approved):

patricklodder:
closing & reopening for CI

chromatic:
One of the CI failures looks like a missing BDB header:

```
In file included from ./wallet/walletdb.h:11:0,
2261                 from ./wallet/wallet.h:19,
2262                 from ./qt/importkeysdialog.h:7,
2263                 from qt/moc_importkeysdialog.cpp:8:
2264./wallet/db.h:21:20: fatal error: db_cxx.h: No such file or directory
2265 #include <db_cxx.h>
2266                    ^
```

The other failure (on Darwin) I believe I've fixed with the latest commit.

patricklodder:
Re: bib failure at https://travis-ci.com/github/dogecoin/dogecoin/jobs/506711074#L2264 - this is because with `NO_WALLET=1` compilation, there is nowhere to store the keys.

I think the problem is the inclusion of `wallet/wallet.h` in `importkeysdialog.h`? I think you can just import that in the .cpp file instead and it be good, since you already have `walletmodel.h` in there?

chromatic:
I like the idea, but it might not be that easy because `importkeysdialog.h` declares `rescanWallet(CWallet*, CBlockIndex*)` as a `Q_SIGNAL`.

I've mentioned both classes in the header file directly. It'd be nice to exclude all of this QT wallet importing code in case of `NO_WALLET=1` but that's not obvious to me right now.

src/qt/bitcoingui.cpp
```diff
@@ -392,6 +393,9 @@ void BitcoinGUI::createActions()
     openAction = new QAction(platformStyle->TextColorIcon(":/icons/open"), tr("Open &URI..."), this);
     openAction->setStatusTip(tr("Open a dogecoin: URI or payment request"));
 
+    importPrivateKeyAction = new QAction(platformStyle->TextColorIcon(":/icons/address-book"), tr("&Import Private Key..."), this);
```
patricklodder, line 396: Non-blocking: Would love to have a new icon for this. Can merge regardless now though, we can change it later too.

src/qt/forms/importkeysdialog.ui
```diff
@@ -0,0 +1,226 @@
...
+           </property>
+           <property name="buddy">
+            <cstring>privateKeyLabel</cstring>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QLineEdit" name="privateKeyLabel">
+           <property name="toolTip">
+            <string>Third party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. %s in the URL is replaced by transaction hash. Multiple URLs are separated by vertical bar |.</string>
```
patricklodder, line 69: This tooltip needs to be changed.

src/qt/forms/importkeysdialog.ui
```diff
@@ -0,0 +1,226 @@
...
+          </widget>
+         </item>
+        </layout>
+       </item>
+       <item>
+        <layout class="QHBoxLayout" name="horizontalLayout_2_ImportPrivateKey">
+         <item>
+          <widget class="QLabel" name="privateKeyLabelWidgetLabel">
+           <property name="toolTip">
+            <string>Third party URLs (e.g. a block explorer) that appear in the transactions tab as context menu items. %s in the URL is replaced by transaction hash. Multiple URLs are separated by vertical bar |.</string>
```
patricklodder, line 56: This tooltip needs to be changed.

src/qt/importkeysdialog.cpp
```diff
@@ -0,0 +1,164 @@
...
+            tr("Invalid address generated from private key; please check and try again!"
+        ));
+        return false;
+    }
+
+    pwalletMain->mapKeyMetadata[vchAddress].nCreateTime = 1;
+
+    if (!pwalletMain->AddKeyPubKey(key, pubkey)) {
+        vchSecret.SetString("");
+        ui->privateKeyImportTextMessage->setText(tr("Couldn't add private key; does it already exist?"));
```
patricklodder, line 126: Are we sure that when a key already exists, we want to react differently than when it doesn't? IIRC this is why RPC doesn't throw an error on duplicate key import, because that could be used to expose which keys are in the wallet?

patricklodder (changes requested):

src/qt/importkeysdialog.cpp
chromatic: Yeah, I'm not sure what the right balance is here. It'd be nice to give some suggestion of what the user could do to correct the problem, but I tried to phrase things vaguely enough that it's not leaking information.

Maybe something more like

  Couldn't add private key. This can be for several reasons:
  
   * the key already exists
   * the key is malformed
   * the wallet is malformed
 
I don't care much for that; it's a lot of text, it's vague, and it still suffers from the flaw you mention. Maybe "Failed to add private key" is the least worst option.

patricklodder: Alright. I'm okay with "Failed to add private key" to unblock this for now. 

I will manually test every scenario I can think of between RPC and UI before merge, if I find any useful improvements we can address that in a subsequent PR.

src/qt/forms/importkeysdialog.ui
```diff
@@ -0,0 +1,226 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>ImportKeysDialog</class>
+ <widget class="QDialog" name="ImportKeysDialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>560</width>
+    <height>440</height>
```
patricklodder, line 10: Height is a bit high at the moment.

src/qt/forms/importkeysdialog.ui
```diff
@@ -0,0 +1,226 @@
...
+  </property>
+  <property name="modal">
+   <bool>true</bool>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout">
+   <item>
+    <widget class="QTabWidget" name="tabWidget">
+     <property name="currentIndex">
+      <number>0</number>
+     </property>
```
patricklodder, line 24: I'd remove these, there's no need for the tab widget as there is only one view. Result:
Screen Shot 2021-05-24 at 02 11 27

src/qt/forms/importkeysdialog.ui
```diff
@@ -0,0 +1,226 @@
...
+          <size>
+           <width>20</width>
+           <height>40</height>
+          </size>
+         </property>
+        </spacer>
+       </item>
+      </layout>
+     </widget>
+    </widget>
```
patricklodder, line 107: matching widget closing tag for `QTabWidget`

src/qt/forms/importkeysdialog.ui
```diff
@@ -0,0 +1,226 @@
...
+        <enum>Qt::Horizontal</enum>
+       </property>
+       <property name="sizeHint" stdset="0">
+        <size>
+         <width>40</width>
+         <height>48</height>
+        </size>
+       </property>
+      </spacer>
+     </item>
```
patricklodder, line 196: Not sure if we need this spacer?

patricklodder (changes requested):
I only have look & feel items left. Everything works well, thank you!!!

Tested on macOS 11.3

src/qt/forms/importkeysdialog.ui
chromatic: I changed it to 200, because that's the smallest I could get it to go locally.

src/qt/forms/importkeysdialog.ui
chromatic: Done, thanks!

src/qt/forms/importkeysdialog.ui
chromatic: Looks like we don't. Good catch.

patricklodder:
This is good for me now. Mind rebase-squashing your commits with `git rebase -i 1.14.4-dev` into one big commit? Then we're good to merge and have our first end-to-end UI improvement in.

Thank you for the work!

chromatic:
Squashed and force pushed.

rnicoll (approved):
Built & tested importing of keys from a remote node, all works as expected in positive scenarios at least.

rnicoll:
Also tested the simple invalid private key scenario. The other failure cases in theory should never occur, and I can't see how to test without mocking parts of the backend, seems extreme.

patricklodder (approved):
ACK! Thank you so much!!!

PR #1865, opened 2021-05-01
Update README.md

MichDe:
Update 
"Early 2015 (approximately a year and a half after release) there will be
approximately 100,000,000,000 coins." 
to 
"Early 2015 (approximately a year and a half after release) there were
approximately 100,000,000,000 coins."
because that was 6 years ago and is now the past. Were makes more sense than will be at this point.


patricklodder (approved):
Good w/ me. Will save travis some compute and not retriever CI for this change.

PR #1904, opened 2021-05-10
Snap Support

slightlyskepticalpotat:
Hello,

I've re-implemented #1360 to make the snap build process download the latest official binaries and verify the gitian signatures instead of build from source, and it seems to work better. Similar to the other pull request, this would also let a wide variety of users install Dogecoin Core with a single command. This has dogecoind, dogecoin-qt, and dogecoin-cli all packed in under the commands `dogecoin-core.d`, `dogecoin-core.qt`, and `dogecoin-core.cli`.

I've tested it under Ubuntu 20.04 and Ubuntu 18.04 so far, and it works perfectly on both.
image
image

Finally, anyone reading this can also test it with `sudo snap install dogecoin-unofficial --channel=edge`. The commands are `dogecoin-unofficial.d`, `dogecoin-unofficial.qt`, and `dogecoin-unofficial.cli` to respectively start the daemon, the qt gui, and the cli. Please only do this in a virtual machine in case of any unforeseen bugs.

### DIFF
--- a/contrib/snap/snapcraft.yaml
+++ b/contrib/snap/snapcraft.yaml
@@ -1,0 +1,68 @@
+name: dogecoin-core # you probably want to 'snapcraft register <name>'
+version: '1.14.4'
+summary: Reference client of Dogecoin, a peer-to-peer digital currency like Bitcoin.
+description: |
+  Dogecoin is a cryptocurrency like Bitcoin, although it does not use SHA256 as its proof of work (POW). Taking development cues from Tenebrix and Litecoin, Dogecoin currently employs a simplified variant of scrypt. MIT licenced.
+
+base: core18
+grade: stable
+confinement: strict
+architectures:
+  - build-on: [amd64, i386, arm64]
+
+apps:
+  d:
+    command: dogecoind
+    plugs: [home, removable-media, network, network-bind]
+    environment:
+    # Override HOME so the datadir is located at ~/snap/dogecoin-core/common/.dogecoin/ instead of ~/snap/dogecoin-core/current/.dogecoin/, and each new version of the snap won't have a different data directory: https://docs.snapcraft.io/environment-variables/7983
+      HOME: $SNAP_USER_COMMON
+  qt:
+    command: qt4-launch dogecoin-qt
+    plugs: [home, removable-media, network, network-bind, desktop, x11, unity7]
+    environment:
+      HOME: $SNAP_USER_COMMON
+      DISABLE_WAYLAND: 1
+  cli:
+    command: dogecoin-cli
+    plugs: [home, removable-media, network]
+    environment:
+      HOME: $SNAP_USER_COMMON
+
+parts:
+  dogecoin-core:
+    plugin: nil
+    override-build: |
+      echo "Downloading files..."
+      wget https://github.com/dogecoin/dogecoin/releases/download/v${SNAPCRAFT_PROJECT_VERSION}/dogecoin-${SNAPCRAFT_PROJECT_VERSION}-${SNAPCRAFT_ARCH_TRIPLET}.tar.gz
+      wget https://github.com/dogecoin/gitian.sigs/archive/refs/heads/master.zip
+      unzip master.zip
+      echo "Verifying signatures..."
+      checksum=$(sha256sum dogecoin-${SNAPCRAFT_PROJECT_VERSION}-${SNAPCRAFT_ARCH_TRIPLET}.tar.gz)
+      if ! grep -r $checksum *; then
+        echo "Signature not verified."
+        return
+      fi
+      tar -xvf dogecoin-${SNAPCRAFT_PROJECT_VERSION}-${SNAPCRAFT_ARCH_TRIPLET}.tar.gz
+      echo "Running tests..."
+      dogecoin-${SNAPCRAFT_PROJECT_VERSION}/bin/test_dogecoin
+      echo "Installing Dogecoin..."
+      install -m 0755 -D -t $SNAPCRAFT_PART_INSTALL/bin dogecoin-${SNAPCRAFT_PROJECT_VERSION}/bin/dogecoind
+      install -m 0755 -D -t $SNAPCRAFT_PART_INSTALL/bin dogecoin-${SNAPCRAFT_PROJECT_VERSION}/bin/dogecoin-qt
+      install -m 0755 -D -t $SNAPCRAFT_PART_INSTALL/bin dogecoin-${SNAPCRAFT_PROJECT_VERSION}/bin/dogecoin-cli
+      wget https://raw.githubusercontent.com/dogecoin/dogecoin/v${SNAPCRAFT_PROJECT_VERSION}/share/pixmaps/bitcoin128.png
+      install -m 0644 -D -t $SNAPCRAFT_PART_INSTALL/share/pixmaps bitcoin128.png
+    build-packages:
+      - unzip
+      - wget
+    stage-packages:
+      - libfreetype6
+      - libpng16-16
+      - libx11-xcb1
+      - libx11-6
+      - libfontconfig
+  qt4conf:
+    build-packages:
+      - dpkg-dev
+    plugin: make
+    source: https://github.com/kyrofa/qt4conf.git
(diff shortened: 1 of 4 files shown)


patricklodder:
I'll have to close & reopen this one more time to deal with 1.14.4-dev CI changes from #1976 - sorry!

contrib/snap/snapcraft.yaml
```diff
@@ -0,0 +1,66 @@
+name: dogecoin-core # you probably want to 'snapcraft register <name>'
+version: '1.14.3'
```
patricklodder, line 2: If we're merging this into 1.14.4-dev, should we name this differently, eg, `1.14.4-pre` or something like that?

contrib/snap/snapcraft.yaml
```diff
@@ -0,0 +1,66 @@
...
+summary: Reference client of Dogecoin, a peer-to-peer digital currency like Bitcoin.
+description: |
+  Dogecoin is a cryptocurrency like Bitcoin, although it does not use SHA256 as its proof of work (POW). Taking development cues from Tenebrix and Litecoin, Dogecoin currently employs a simplified variant of scrypt. MIT licenced.
+
+base: core18
+grade: stable
```
patricklodder, line 8: Not sure about "stable" for development branch either?

patricklodder (commented):
I have 2 questions about the `snapcraft.yaml` values. Would appreciate thoughts.

contrib/snap/snapcraft.yaml
slightlyskepticalpotat: Ah, I don't think I was clear enough in my explanation. This only works for actual releases because it grabs the releases from the [releases page](https://github.com/dogecoin/dogecoin/releases), where there are files named `dogecoin-1.14.3-...`. If we intended to release this in 1.14.4, it would need to be versioned 1.14.4 and would only start working when the 1.14.4 release is out.

contrib/snap/snapcraft.yaml
slightlyskepticalpotat: As mentioned in the above comment, this only works for reasons that have a formal github release. There are [four grades](https://snapcraft.io/docs/channels), and I thought stable fit the best.

patricklodder:
> [..] If we intended to release this in 1.14.4, it would need to be versioned 1.14.4 and would only start working when the 1.14.4 release is out.

Alright! I'll resolve my comments. If you could please rebase off 1.14.4-dev and resolve the `.gitignore` conflict, I'll look into updating release documentation!

Thanks!!!

slightlyskepticalpotat:
Rebased and resolved, I'm going to build and test the package again to be safe.

slightlyskepticalpotat:
Once again working on my system, this version should automatically work with the 1.14.4 release once that is released.

contrib/snap/snapcraft.yaml
```diff
@@ -0,0 +1,68 @@
...
+  dogecoin-core:
+    plugin: nil
+    override-build: |
+      echo "Downloading files..."
+      wget https://github.com/dogecoin/dogecoin/releases/download/v${SNAPCRAFT_PROJECT_VERSION}/dogecoin-${SNAPCRAFT_PROJECT_VERSION}-${SNAPCRAFT_ARCH_TRIPLET}.tar.gz
+      wget https://github.com/dogecoin/gitian.sigs/archive/refs/heads/master.zip
+      unzip master.zip
+      echo "Verifying signatures..."
+      checksum=$(sha256sum dogecoin-${SNAPCRAFT_PROJECT_VERSION}-${SNAPCRAFT_ARCH_TRIPLET}.tar.gz)
+      if ! grep -r $checksum *; then
```
rnicoll, line 42: Can we create a variable for the expected checksum, please, so it's a bit more obvious it will need filling in at release time?

rnicoll: That said; realising this creates a chicken-egg problem in that it needs filling in after the release is tagged, doesn't it? Hmm. Probably worth calling out in the docs if we expect end users to use this.

rnicoll:
I think generally looks good, I'd like a discussion on whether we want it targetting 1.14.4 and having a blank checksum, or use 1.14.3 and a defined checksum. In theory at least anyone preparing the snap MUST edit the file to insert the checksum, so I'd not be concerned about it having the previous version in.

contrib/snap/snapcraft.yaml
slightlyskepticalpotat: If we fill in the expected checksum, we would only be able to target version 1.14.3 with this. Is there a way we could make the current system more robust? Currently, it checks if the checksum generated exists anywhere within the gitian repo, which should be updated at each release.

edtubbs:
Checksum for current (stable/test) embedded in a single script for both documentation and verification is good; they're also published on the site for independent verification.

fmhc:
Got it installed on a Ubuntu VM and its running, but I get some warnings / errors from the Fontconfig and also got an xkbcommon error, don't know if this has to do with me installing ubuntu server + adding gnome ontop later but i guess it could also be something QT-related. I will try again with a completely fresh ubuntu desktop VM later
image
image

contrib/snap/snapcraft.yaml
fmhc: I think we should have the checksums in a specific file in a specific branch and not grep over the whole repository, just to close the gap that someone _could_ exploit it with a hash somewhere else?

edtubbs:
"xkbcommon: ERROR: failed to add default include path error" may have been addressed in Qt.

https://github.com/conda-forge/qt-feedstock/issues/104

fmhc:
> "xkbcommon: ERROR: failed to add default include path error" may have been addressed in Qt.
> 
> [conda-forge/qt-feedstock#104](https://github.com/conda-forge/qt-feedstock/issues/104)

yes, i can open a wallet but i'm not able to type in anything. Are there tests covering this?

slightlyskepticalpotat:
I was able to type in and interact with it normally on a fresh Ubuntu 20.04.2 desktop install with the English (US) keyboard. Could it be something to do with German specifically or non-english languages?

contrib/snap/snapcraft.yaml
slightlyskepticalpotat: There are no public preimage or collision attacks on full SHA-256 right now, so that would be quite the accomplishment.

edtubbs:
Agreed, Qt should probably be at 5.15 at least.

rnicoll:
Is it worth upgrading Qt or better to wait for 1.21 to bring it in, what do people think?

If we do need it, please raise an issue.

Beyond that, I'm still in favour of coding in 1.14.3's checksum, and expect people using this to package a release, to be able to replace the values. Means if anyone gets it wrong, they're still safe, it can't be a hacked version. It would need to run GPG to actually verify the signature, as an alternative.

fmhc:
> I think generally looks good, I'd like a discussion on whether we want it targetting 1.14.4 and having a blank checksum, or use 1.14.3 and a defined checksum. In theory at least anyone preparing the snap MUST edit the file to insert the checksum, so I'd not be concerned about it having the previous version in.

I think having an initial defined checksum is a good starting point.

> Checksum for current (stable/test) embedded in a single script for both documentation and verification is good; they're also published on the site for independent verification.

having it in a seperate file would be good from a security perspective, to parse just one file for that. 
one could go down the rabbit hole and think about storing hashes on dogecoin, as messages from devs to a specific wallet or sth. - but then we would really have a chicken-egg problem 😹

contrib/snap/snapcraft.yaml
fmhc: i like it the way they did it here http://qt-mirror.dannhauer.de/archive/qt/5.15/5.15.2/submodules/md5sums.txt

fmhc:
I've created https://github.com/dogecoin/dogecoin/issues/2338 for the keyboard issue with Qt, hope that this is the correct workflow. Have forked and cloned / switched to v1.14.4-dev, created a new branch and now I try to get it work with a newer Qt version -> move discussion about that in the issue i guess ;)

rnicoll (approved):
Going to merge this now and refine to a single SHA256 sum for security.

PR #1910, opened 2021-05-11
Rename to caticoin

Jolly-Pirate:


Jolly-Pirate:
Oops, made the wrong PR.

PR #1946, opened 2021-05-16
Update QT client messages and translations to Chinese Doge equivalents.

oiooj:
Replace "比特币" （Bitcoin in Chinese） with “狗狗币” （Dogecoin in Chinese）, the `qt-ui` is so important for community users.

We still have a lot of work to do in this field.

/cc @rnicoll


slightlyskepticalpotat:
Please see this, which fixes those lines and more: https://github.com/dogecoin/dogecoin/pull/1871

PR #1953, opened 2021-05-16
1.21 block interval regtest

rnicoll:
Follow-on to #1780 which applies block interval changes to the regression test network.


rnicoll:
This has passed Travis on my own branch, but failed on the PR: https://travis-ci.com/github/rnicoll/dogecoin

Rather than forcing a re-run, I'll rebase once #1780 goes in, and resolve any issues after.

l4p4:
Ok, nice change! Just adjust conflicts file

rnicoll:
In the end rolled this into #1780

PR #1959, opened 2021-05-16
Fixing travis.yml build errors - replaced trusty with bionic in .travis.yml

xtemp3r:
[Build Status](https://travis-ci.com/xtemp3r/dogecoin)

changing the distribution from trusty to bionic so it fixed 5 out of 7 build errors because of it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: bionic
 os: linux
 language: generic
 cache:


urbfigviera (approved):

rnicoll:
That does seem to help, although have we then got version compatibility issues? Seeing things like:

```
net.h:397:10: error: ‘mutex’ in namespace ‘std’ does not name a type

     std::mutex mutexMsgProc;
```

That said, given I build on 20.04 personally, a bit confused. Will come back to this, want to get a second opinion.

rnicoll:
Okay, the two failing builds are mingw builds (i.e. Windows), and appear to relate to C++11 features not available in that environment:

* https://en.cppreference.com/w/cpp/thread/condition_variable
* https://en.cppreference.com/w/cpp/thread/mutex
* https://en.cppreference.com/w/cpp/thread/thread

As this is minor to revert, and makes the majority case better, merging.

rnicoll (approved):

rnicoll:
Thank you!

soljarjb13 (commented):
V

soljarjb13 (commented):
V

soljarjb13 (commented):
V

PR #1976, opened 2021-05-20
Travis: switch back to trusty and remove useless repos

langerhans:
The mingw compiler in bionic seems to be having the same issues that xenial had. Given the actual reason our builds failed with trusty was just a 404 in pgsql's apt repo the simple fix is just not querying these repos. As a precaution cause other vendors might be removing their trusty repos, all 3rd party sources are removed before apt is called.

This has the added advantage that travis stays in line with gitian for this branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: bionic
+dist: trusty
 os: linux
 language: minimal
 cache:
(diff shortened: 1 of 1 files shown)


patricklodder (approved):
Works!!! Thank you ❤️

PR #2015, opened 2021-05-24
Dogecoin FIXED

vicepool:
Too many fixes to describe. 
2 119 changes files.
974 deletions.
Unnecessary original files moved to the appropriate folder.
Works better than "10x" approach suggested.


rnicoll:
Huh, well at least you didn't just delete it all.

PR #2065, opened 2021-05-25
Code Signing Doge Releases with Pub/Private Keys created from Sound of [Universe + Moon + Earth Tides + SpaceX Raptor] and measurements of [Earth to Moon Distance + Mars Station Sqft + Tesla Roadster Travel Distance] and Doge Community Signatures

satscreate:
**Have you ever code signed a distribution of Doge.** If not 
Here is my crazy idea to generate the Universe Strong cryptographic Key creation, which makes doge distribution 

1. Ensure that a piece of code has not been altered since it was signed.
2. Identify code as coming from a specific source (a developer or signer).
3. Determine whether code is trustworthy for a specific purpose.

There are some factors to consider while creation of public private key pairs for the code signing, which makes difficult to crack by this universe at least.

1. Sound of [Universe + Moon + Earth Tides + SpaceX Raptor]  (Dynamic)
2. Measurements of [Earth to Moon Distance + Mars Station Sqft + Tesla Roadster Travel Distance]  (Dynamic)
3. Doge Community Signatures (Dynamic)

Thanks for giving people an opportunity to be a part of peoples currency. Doge to the Moon. :)


rnicoll:
I like to write something constructive before closing but I have no good way to tell you this isn't practical to implement.

Issue #2009, opened 2021-05-24
[bug] 1.21 Travis timeouts


rnicoll:
Currently the time to complete tests on CI (Travis) is too long, or the timeouts too short, or potentially both.

We need to identify a path to maintaining test quality while identifying opportunities to improve reliability of the tests. A paid Travis subscription is one potential option, but this would still mean others would struggle to get Travis to pass on personal repos.

Acceptance criteria:

* Where CI passes successfully on system without timeout (developer machine), it passes on Travis, excluding infrastructure failures.

mmicael1:
@rnicoll need help? might be a good idea to dockerize it!

rnicoll:
@mmicael1 It's Dockerized, take a look at https://github.com/dogecoin/dogecoin/tree/1.21-dev/ci/test . I thought the problem was that for Dogecoin we have faster blocks, so we have more blocks for coinbase maturity, so as a result we need more blocks when doing tests.

I had a trial on https://github.com/rnicoll/dogecoin/tree/1.21-travis-timeouts of using more flexible coinbase maturity values, however for the functional tests it seemed to make no difference at all. I might re-evaluate with a focus on the unit tests (under https://github.com/dogecoin/dogecoin/tree/master/src/test ) though, perhaps I was looking in the wrong place.

mmicael1:
IMO, you should 1st create a pre-image already built which will save about 5mn of install.  (cf https://travis-ci.com/github/dogecoin/dogecoin/jobs/507022866)
Also maybe it would be better to use Dockerfile instead of having to install local dependencies on computers and benefits of caching.
Currently I m not even able to launch it locally (I’m sorry, 'getopt --test' failed in this environment. Please load GNU getopt.
) 😢 
I know I have a MAC :trollface:
Maybe it would be the 1st step for a lot of new developers to help here...

rnicoll:
@mmicael1 Is that something you can tackle trying? Although eying https://travis-ci.com/github/dogecoin/dogecoin/jobs/507937592 and thinking maybe I've got it down to a single unstable test, and if I can fix that, a bigger change would be unnecessary, so if it's likely to be a massive time sink for you, maybe see if I can fix that test instead.

ashu0801:
Include me in the Doge development as well.. I want to contribute and having 7 years of working experience as a software engineer

mmicael1:
[...]

PR #2115, opened 2021-05-25
Sync nodes before maturing blocks in bumpfee

rnicoll:
In Dogecoin we have a much higher coinbase maturity value, so we need a lot more blocks after a block was mined, to be able to spend it. In `wallet_bumpfee` I solved this by having another node mine after the node which receives the blocks, however as the network doesn't sync before changing which node is mining, potentially this is leading to a race condition.

Opening this as a potential solution, but primarily to verify on Travis.

### DIFF
--- a/test/functional/wallet_bumpfee.py
+++ b/test/functional/wallet_bumpfee.py
@@ -66,6 +66,7 @@ class BumpFeeTest(BitcoinTestFramework):
         # fund rbf node with 10 coins of 0.001 btc (100,000 satoshis)
         self.log.info("Mining blocks...")
         peer_node.generate(110)
+        self.sync_all()
         # Dogecoin: Coinbase maturity is much higher, so we mine to the RBF node for funding
         rbf_node.generate(140)
         self.sync_all()


angelcervantesa:
hi, hope all are having a good day. Can I investigate this further as well?

rnicoll:
@angelcervantesa Yes, although suggest looking through the issue I've raised as there's a bundle of test issues I'm trying to fix: https://github.com/dogecoin/dogecoin/issues/2009

angelcervantesa:
> @angelcervantesa Yes, although suggest looking through the issue I've raised as there's a bundle of test issues I'm trying to fix: https://github.com/dogecoin/dogecoin/issues/2009
> 
> 
got it. will get back to this after i get rid of this cold.

patricklodder (approved):
👍

angelcervantesa:
> > @angelcervantesa Yes, although suggest looking through the issue I've raised as there's a bundle of test issues I'm trying to fix: #2009
> 
> got it. will get back to this after i get rid of this cold.

Currently still coughing, but was about to clone repo, investigating further.

Happy Tuesday <3

Issue #2138, opened 2021-05-25
In a Windows environment, running make HOST=x86_64-w64-mingw32 throws various syntax errors


jimmyt1988:
# Bug Report

**Current behavior:**

When following the `doc/build.windows.md` on Windows 10, if you run `make HOST=x86_64-w64-mingw32`, the command prompt throws the following error:

```txt
: not foundb: 4:
: not foundb: 6:
: not foundb: 26:
: not foundb: 27:
: not foundb: 34:
: not foundb: 37:
: not foundb: 45:
: not foundb: 52:
: not foundb: 54:
: not foundb: 66:
: not foundb: 74:
: not foundb: 77:
./config.sub: 80: Syntax error: word unexpected (expecting "in")
: not foundb: 4:
: not foundb: 6:
: not foundb: 26:
: not foundb: 27:
: not foundb: 34:
: not foundb: 37:
: not foundb: 45:
: not foundb: 52:
: not foundb: 54:
: not foundb: 66:
: not foundb: 74:
: not foundb: 77:
./config.sub: 80: Syntax error: word unexpected (expecting "in")
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Makefile: not found
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
/bin/sh: 1: Syntax error: "|" unexpected
make: config.site.in: Command not found
[...]

PR #2140, opened 2021-05-25
Added additional information to help new developers navigate first builds in Windows

jimmyt1988:
I have added an additional piece of information to the documentation to assist new developers like myself with building for the first time.

Related issue: #2138

### DIFF
--- a/doc/build-windows.md
+++ b/doc/build-windows.md
@@ -96,6 +96,9 @@ results in `Autoconf` configure scripts being able to execute Windows Portable E
 unexpected behaviour during the build, such as Win32 error dialogs for missing libraries. The recommended approach
 is to temporarily disable WSL support for Win32 applications.
 
+Note that if you are are running into the following errors while running `make HOST=x86_64-w64-mingw32`, ensure that your Line endings are set to LF (Linux) and not CLRF (Windows).
+> ./config.sub: 80: Syntax error: word unexpected (expecting "in")
+
 Build using:
 
     PATH=$(echo "$PATH" | sed -e 's/:\/mnt.*//g') # strip out problematic Windows %PATH% imported var


doc/build-windows.md
```diff
@@ -96,6 +96,9 @@ results in `Autoconf` configure scripts being able to execute Windows Portable E
 unexpected behaviour during the build, such as Win32 error dialogs for missing libraries. The recommended approach
 is to temporarily disable WSL support for Win32 applications.
 
+Note that if you are are running into the following errors while executing `make HOST=x86_64-w64-mingw32`, ensure that your Line endings are set to LF (Linux) and not CRLF (Windows) on the `config.guess` and `config.sub` files found in the depends folder.
+> ./config.sub: 80: Syntax error: word unexpected (expecting "in")
```
rnicoll, line 100: Instead of `> ` for formatting, maybe using backticks, to make it clear it's in-line code. See also https://meta.stackexchange.com/questions/189920/what-is-the-3-x-backquote-markdown-used-for

jimmyt1988: Thanks for the input! I think I'm missing something here :) - Allow me to be dumb for a second hehe: this is currently how the edit looks:

image

Would you prefer it look like this?

image

If so. Cool :)

rnicoll: I was thinking (B)

jimmyt1988: B it is. I've also squashed. Thanks!

AbcSxyZ:
[Do not have windows to test anything].
Some suggestions in order to improve the PR.

By reading the text, it was unclear where the problem came from. It's not a problem related to `config.guess` and `config.sub` files, but to every files because it's a `git` misconfiguration, as you mentionned in #2138. In the PR, you don't mention git and do not explain how to solve the issue, some research will be necessary.

I think **instead of speaking about potential errors, it would be more interesting to add instruction to avoid them.**

You suggest to change global git settings, I do not think it's a good advice because it could break user config for other repo, could create some trouble for him later.

There is probably something to do with [`.gitattributes` text setting](https://git-scm.com/docs/gitattributes#_text).
If I understand correctly, git on windows could convert automatically EOL from LF to CRLF, which create trouble, and regarding dogecoin Windows documentation a Linux style EOL seem expected.

If you add the following in `.gitattributes`, does it solve the problem ?
```
*    text eol=lf
```
It force commit & clone to use LF.

Hope it won't breaking anything, can be dangerous and I don't master the setting. If it work for you, need some check on other environments.

jimmyt1988:
> If you add the following in `.gitattributes`, does it solve the problem ? _[Re-clone would be needed]_
> 
> ```
> *    text eol=lf
> ```

Interestingly I can't get this to work yet. Who would have thought trying to force a file to both save and clone as `LF` would be so blooming difficult. I'll have another crack at this on Monday evening. I have tried the following variations to no avail and have ensured that my git config has no auto crlf options enabled:

- 1
  - Save CRLF file to a repository
  - Add .gitattributes with `*    text eol=lf`
  - Re-clone repository.
- 2
  - Save CRLF file to a repository
  - Add .gitattributes with `*.txt eol=lf`
  - Re-clone repository
- 3
  - Steps 2 and 3 copied but saving an LF to the repository instead.

Let's hope I crack this later haha. Sheesh.

AbcSxyZ:
Take a look at my [git-testing repo](https://github.com/AbcSxyZ/git-testing). It's something minimalist to test the feature, I'm able to create LF file, which are converted to CRLF when pushed. If it can help you to figure out what's going wrong.

Switch the instruction order, if you save a CRLF file to a repo, then add `.gitattributes` instruction, the file will still be CRLF on the repo and on clone. It seems you have to `git add` the file after changing `.gitattributes` to have the change to take effect.

> Add .gitattributes with `*.txt    eol=lf`

The exact syntax is `pattern attr1 attr2 ...`, it's possibly a wrong instruction, should be `*.txt    text    eol=lf`.

jimmyt1988:
> Take a look at my [git-testing repo](https://github.com/AbcSxyZ/git-testing). It's something minimalist to test the feature, I'm able to create LF file, which are converted to CRLF when pushed (and vice-versa). If it can help you to figure out what's going wrong.
> 
> Switch the instruction order, if you save a CRLF file to a repo, then add `.gitattributes` instruction, the file will still be CRLF on the repo and on clone. It seems you have to `git add` the file after changing `.gitattributes` to have the change to take effect.
> 
> > Add .gitattributes with `*.txt    eol=lf`
> 
> The exact syntax is `pattern attr1 attr2 ...`, it's possibly a wrong instruction, should be `*.txt text eol=lf`.

@AbcSxyZ , @rnicoll 

After having a play around with this (check out the stream here: https://www.youtube.com/watch?v=Z2wucCAlQaU), I have:
- Gone through the `Makefile` and included the files that are processed by the build, inside the `.gitattributes` file. 
- I've tested a new `git clone` with those above changes and it has corrected the issue of `CRLF`. I've tested back and forth a few times to ensure I've hit the right nails.
- I've removed the comments I originally made in the build-windows.md as the inclusion of the `.gitattributes` file has addressed the problem at source.
- I've run the make command mentioned in the windows.build and it is now functioning correctly.

Thanks so much in advance!

.gitattributes
```diff
@@ -1 +1,8 @@
 src/clientversion.cpp export-subst
+
+# Ensures Windows pulls line ending dependent files correctly for Makefile's build process:
+depends/Makefile text eol=lf
```
AbcSxyZ, line 4: If it works by explicitly include for all files, a wildcard should do the work.

Replacing all of those lines by `*    text    eol=lf` should be equivalent, forcing the entire repo to use always use LF and avoid CRLF line endings.

jimmyt1988: We'll probably want to wildcard only files being built in the makefile. I wouldn't want to LF the visual studio solution, for example.

AbcSxyZ (changes requested):
Great improvement to make a more appropriate solution :) Nice work, also a great initiative to make some dogecoin development video [although if fixing line ending shouldn't be the most exciting part :smile:] !

## To solve before potential merging

I suggested changing the instruction by a wildcard to include all the files, it will make the line ending consistent for the entire repo (use of `* text eol=lf`).

A potential issue I see where we need feedback from core devs is : **would some users expect CRLF ending ? Would it be a problem to force dogecoin line ending to always be in a Unix style ?**

If I'm not wrong, adding this instruction shouldn't make any trouble to unix users, but would create some for a Windows user who would expect CRLF line ending.

The current Windows docs suggest to use Windows Subsystem for Linux, and nothing in the installation seem adapted specifically for Windows machine. In this case, seem possible from my pov to force LF use.

jimmyt1988:
> The current Windows docs suggest to use Windows Subsystem for Linux, and nothing in the installation seem adapted specifically for Windows machine. In this case, seem possible from my pov to force LF use.

Yes, for the case of WSL building. But don't forget the Visual Studio specific folder which also acts as a build mechanism - https://github.com/dogecoin/dogecoin/tree/1.21-dev/build_msvc - we just need to be careful around stepping on any unnecessary files.

AbcSxyZ:
So if VS Code need CRLF on Windows to perform build, it's definitively a problem. IDK the software but because it's cross-platform, possibly it's agnostic of line endings and compile with both. Some confirmations are needed.

> we just need to be careful around stepping on any unnecessary files.

But being general is also easier to maintain & more resilient to changes. It can also be great to be consistent for the entire repo, giving the user only LF or CRLF files, I'm not sure mixing both is the best way [but perhaps it is].

There is probably some `.gitattributes` good practices who address this issue.

jimmyt1988:
> So if VS Code need CRLF on Windows to perform build, it's definitively a problem. 

Just to ensure we're on the same page. "Visual Studio" not "Visual studio code". They are very different programs hehe. Visual studio is not cross platform in the way Visual studio code is.

> But being general is also easier to maintain & more resilient to changes. 

Let's see what the consensus is among others. I'm more for changes being specific to the matters they address, but I'm happy to be outvoted :-).

> It can also be great to be consistent for the entire repo, giving the user only LF or CRLF files, 

It really depends. If files are specifically designed for Windows, like the Visual Studio solution and its working parts, then I don't see the benefit to convert to LF regardless of the rest of the files being LF.

> Possibly some `.gitattributes` good practices who address this issue.

I had noted in the YouTube video a few stack overflow posts suggesting to be specific. However, let's see what others say :-)

The main goal to keep in mind is that this is the first step a Windows developer gets to, and the error is not obvious, so we should ensure it is obvious from now on... or that it doesn't happen.

rnicoll:
Can someone developing on Windows confirm this makes things better for them, please? It looks good from a cursory read through just... y'know... I'd like someone who's not primarily a Linux dev to check for me.

jimmyt1988:
> Can someone developing on Windows confirm this makes things better for them, please? It looks good from a cursory read through just... y'know... I'd like someone who's not primarily a Linux dev to check for me.

Yes please. Haha. Great idea. Thanks for both of your input. Can't wait to see what improvements I can make!!!

patricklodder:
I'm untagging this from 1.14.4, putting in to the 1.14.5 project and tagging with `help needed`

PR #2154, opened 2021-05-26
depends: update Qt 5.9 source url

rnicoll:
Backport a fix from BC to correct the Qt download URL for CI. It worked without this, but we were grabbing files from Bitcoin's own mirror, rather than the official site.


jimmyt1988:
Looks reasonable to me

patricklodder (approved):

kitcateleven (commented):
Sweet!****

kitcateleven (commented):
Looks great
like a (package)_version

PR #2203, opened 2021-05-29
1.21 Update min chain work and assume valid

rnicoll:
Update minimum chain work and block hash to assume preceeding blocks are valid, for main and test networks.


patricklodder (approved):
LGTM.

PR #2204, opened 2021-05-29
Replace feature enablement blocks

rnicoll:
Replace block hash/heights when features are enabled.


metatron1973:
Nice teamwork here.

patricklodder (approved):
Those look good. 

Note: I don't like the `BIP16Exception` because it's Bitcoin specific, but this is a good solution for now, we'll clean it up later.

PR #2221, opened 2021-05-30
Update DB version to 5.1

rnicoll:


bjacquin:
Could you please consider raising to berkeleyDB 5.3 as 5.1 is being deprecated from distro like Gentoo (https://bugs.gentoo.org/792222) and Fedora (https://fedoraproject.org/wiki/Changes/Libdb_deprecated)

haxpor:
also similarly apply to Ubuntu 20.04, by default I only see 5.3, and no 5.1 there.

michilumin:
Looking into this, in re: 5.3, as I've run into the same problems with MacOS and homebrew not wanting to play nice with < 5.3.  Acknowledged and looking into it.

michilumin:
so it looks like Ross is all good to update this to 5.3 after I do some compatibility checks on 4.8 -> 5.3 forward compatibility, and 5.3-created DBs going at least back to 5.1.   Recently while doing some wallet recoveries for people i've seen decent interop between 5.1 and 5.3; and i know other coins are using 6.x without trouble, so we **should** be OK but I want to verify first.

rnicoll:
Closing as replaced by #2245

PR #2222, opened 2021-05-30
Articols matter in Italian

rnicoll:


Klian326 (approved):

Padanian (changes requested):
Troll pull request. To be deleted.

rnicoll:
@Padanian it's a forward port of a suggested change introduced in 1.14, to 1.21. Can you be more specific about the problem with it? Although fully admit I'm taking the commit on trust

Padanian:
The title of the pull is a mispronounced translitteration of the italian word "Articolo" which stands for the "Definite article". Then, nouns in Italian do not require the capital letter, which is only required in German nouns. Finally, the former translation is perfectly understandable, and is common practice in software engineering or user interfaces to omit the definite article.

Klian326:
The translation is correct in both ways, I can't understand why you're pointing this out and you don't just approve it. To me, it is just someone from 2017 trying to improve the italian translation, not a troll.

Padanian:
It's not wrong in the strict term, it's just useless, tipical trolling, definitely not improving anything.

Klian326:
Yeah, I agree, that's why I approved and moved on... I mean, you're right and this could be avoided but at this point let's close this commit as soon as possible and let's move on lol
Also, I'm a Padanian too hahaha

src/qt/locale/bitcoin_it.ts
```diff
@@ -89,7 +89,7 @@ E' possibile firmare solo con indirizzi di tipo "legacy".</translation>
     </message>
     <message>
         <source>Export Address List</source>
-        <translation>Esporta elenco degli indirizzi</translation>
+        <translation>Esporta la Lista degli Indirizzi</translation>
```
virginiavirduzzo, line 92: Use downcase for those words: `lista` & `indirizzi`.
Should be like: `Esporta la lista degli indirizzi`

rnicoll:
```suggestion
        <translation>Esporta la lista degli indirizzi</translation>
```

patricklodder (approved):

PR #2224, opened 2021-05-31
fix: reduce tests latency

mmicael1:
Non-necessary long time sleep for sync_blocks and sync_mempools.


rnicoll:
Thanks, looking at merging this after #1828

rnicoll (approved):
ACK, tested on my box and works well.

PR #2243, opened 2021-06-04
Qt: Add sortable sent/recv bytes to Peers debug table

chromatic:
See GH #2240.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2011-2016 The Bitcoin Core developers
+// Copyright (c) 2021 The Dogecoin Core developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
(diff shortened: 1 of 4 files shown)


src/qt/guiutil.cpp
```diff
@@ -941,6 +942,11 @@ QString formatPingTime(double dPingTime)
     return (dPingTime == std::numeric_limits<int64_t>::max()/1e6 || dPingTime == 0) ? QObject::tr("N/A") : QString(QObject::tr("%1 ms")).arg(QString::number((int)(dPingTime * 1000), 10));
 }
 
+QString formatUintValue(uint64_t uValue)
+{
+    return (uValue == std::numeric_limits<int64_t>::max()/1e6 || uValue == 0) ? QObject::tr("N/A") : QString(QObject::tr("%1")).arg(QString::number(uValue), 10);
```
chromatic, line 947: Review welcome on the experience here; it'd be nice to turn this into KB, MB, GB, etc... but there could be some Qt feature I don't know of that formats values nicely here.

rnicoll: I'd typically write it over a few more lines for simplicity, but... my inclination is there's some really clever ways we could do this, but I'm happy with anything that works.

rnicoll (approved):
LGTM.

I do wonder if we should make the window default size bigger, and I'd like a second review from someone with more Qt experience, but tested and it seems to work.

src/qt/guiutil.cpp
chromatic: Qt 5.10 introduced [formattedDataSize](https://doc.qt.io/qt-5/qlocale.html#formattedDataSize-1), but I'm not sure we require Qt 5 fully, so I might just write a 10-line loop instead.

src/net.h
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2016 The Bitcoin Core developers
+// Copyright (c) 2021 The Dogecoin Core developers
```
patricklodder, line 3: seems to be no other change than the copyright message?

src/qt/guiutil.cpp
patricklodder: Recommend to do the 10-line loop for 1.14.4 because technically we still support Qt4

src/qt/peertablemodel.cpp
```diff
@@ -114,7 +119,7 @@ PeerTableModel::PeerTableModel(ClientModel *parent) :
     clientModel(parent),
     timer(0)
 {
-    columns << tr("NodeId") << tr("Node/Service") << tr("User Agent") << tr("Ping");
+    columns << tr("NodeId") << tr("Node/Service") << tr("User Agent") << tr("Ping") << ("Bytes Sent") << ("Bytes Received");
```
patricklodder, line 122: Could you please use `tr()` so we can translate this?

patricklodder (changes requested):
I'd like to see display in another unit than bytes, as [it currently already becomes rather unreadable with megabytes of data.](https://user-images.githubusercontent.com/1410115/120855963-6e7aeb80-c57f-11eb-8dc2-09ad4f1580d9.mov)

I'm ok with either fixating on megabytes, or using a dynamic display.

src/net.h
chromatic: That's right; I was over-eager applying a macro to all of my open files. Removed.

src/qt/guiutil.cpp
chromatic: Done, but it's not pretty. I could be persuaded to forget the string-handling format code and create the output directly in each if/else branch. What's the least unmaintainable approach here?

rnicoll: Personally I like if-else chains, they're not pretty, but everyone know what they do.

rnicoll:
Trying this now. Nowish.

rnicoll:
Code works. I'll leave Patrick to verify he's happy with the code itself, but LGTM.

pr

src/qt/guiutil.cpp
```diff
@@ -941,6 +942,38 @@ QString formatPingTime(double dPingTime)
...
+{
+    // Why handle these comparisons directly, instead of a clever algorithm?
+    // This is likely to be called in a tight loop, so avoid the overhead of
+    // setting up a constant list and walking an iterator.
+    static const uint64_t TERABYTE_SIZE = UINT64_C(1024*1024*1024*1024);
+    static const uint64_t GIGABYTE_SIZE = UINT64_C(1024*1024*1024);
+    static const uint64_t MEGABYTE_SIZE = UINT64_C(1024*1024);
+    static const uint64_t KILOBYTE_SIZE = UINT64_C(1024);
+
+    QString unitFormat = QObject::tr("%1");
```
patricklodder, line 955: Can we make this formatting mask "%1 B"?

patricklodder (changes requested):
I have one minor improvement left, then I think this is good to go.

src/qt/guiutil.cpp
chromatic: Done, and squashed everything into a single logical commit.

patricklodder (approved):
ACK. Tested on macOS for both testnet and mainnet, including during sync.

PR #2253, opened 2021-06-05
Add python3-devel package to CentOS

rnicoll:
* Add python3-devel package to CentOS as it is required for ltc_scrypt to build.
* Remove python3-distutils package from CentOS as they ship in the default Python 3 install.


michilumin (approved):
This seems OK for the current Python 3.x development package on RHEL/CentOS 7.

PR #2259, opened 2021-06-05
Add node context to wallet RPC request object

rnicoll:
Add node context to wallet RPC request object, as `getauxblock` needs access to the node to determine if it is in initial block download, and requires access to the wallet to get an address to pay to.

### DIFF
--- a/src/wallet/context.h
+++ b/src/wallet/context.h
@@ -10,6 +10,8 @@ namespace interfaces {
 class Chain;
 } // namespace interfaces
 
+struct NodeContext;
+
 //! WalletContext struct containing references to state shared between CWallet
 //! instances, like the reference to the chain interface, and the list of opened
 //! wallets.
@@ -23,6 +25,9 @@ class Chain;
 struct WalletContext {
     interfaces::Chain* chain{nullptr};
     ArgsManager* args{nullptr};
+
+    //! Dogecoin: getauxwork is a wallet RPC but actually needs the NodeContext.
+    NodeContext* nodeContext{nullptr};
 
     //! Declare default constructor and destructor that are not inline, so code
     //! instantiating the WalletContext struct doesn't need to #include class
--- a/src/wallet/context.h
+++ b/src/wallet/context.h
@@ -5,12 +5,12 @@
 #ifndef BITCOIN_WALLET_CONTEXT_H
 #define BITCOIN_WALLET_CONTEXT_H
 
+#include <node/context.h>
+
 class ArgsManager;
 namespace interfaces {
 class Chain;
 } // namespace interfaces
-
-struct NodeContext;
 
 //! WalletContext struct containing references to state shared between CWallet
 //! instances, like the reference to the chain interface, and the list of opened
--- a/src/wallet/init.cpp
+++ b/src/wallet/init.cpp
@@ -107,7 +107,7 @@ void WalletInit::Construct(NodeContext& node) const
         LogPrintf("Wallet disabled!\n");
         return;
     }
-    auto wallet_client = interfaces::MakeWalletClient(node, *node.chain, args);
+    auto wallet_client = interfaces::MakeWalletClient(node, args);
     node.wallet_client = wallet_client.get();
     node.chain_clients.emplace_back(std::move(wallet_client));
 }
--- a/src/wallet/init.cpp
+++ b/src/wallet/init.cpp
@@ -107,7 +107,7 @@ void WalletInit::Construct(NodeContext& node) const
         LogPrintf("Wallet disabled!\n");
         return;
     }
-    auto wallet_client = interfaces::MakeWalletClient(*node.chain, args);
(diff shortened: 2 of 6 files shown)


himontee:
someone review

src/wallet/context.h
```diff
@@ -10,6 +10,8 @@ namespace interfaces {
 class Chain;
 } // namespace interfaces
 
+struct NodeContext;
+
```
sujayVittal, line 14: Unnecessary line. Could be removed.

src/wallet/test/init_test_fixture.cpp
```diff
@@ -10,7 +10,7 @@
 
 InitWalletDirTestingSetup::InitWalletDirTestingSetup(const std::string& chainName) : BasicTestingSetup(chainName)
 {
-    m_wallet_client = MakeWalletClient(*m_chain, *Assert(m_node.args));
+    m_wallet_client = MakeWalletClient(m_node, *m_chain, *Assert(m_node.args));
```
sujayVittal, line 13: Shouldn't it be an address to the m_node that needs to be passed here?

src/wallet/test/wallet_test_fixture.h
```diff
@@ -21,7 +21,7 @@ struct WalletTestingSetup : public TestingSetup {
     explicit WalletTestingSetup(const std::string& chainName = CBaseChainParams::MAIN);
 
     std::unique_ptr<interfaces::Chain> m_chain = interfaces::MakeChain(m_node);
-    std::unique_ptr<interfaces::WalletClient> m_wallet_client = interfaces::MakeWalletClient(*m_chain, *Assert(m_node.args));
+    std::unique_ptr<interfaces::WalletClient> m_wallet_client = interfaces::MakeWalletClient(m_node, *m_chain, *Assert(m_node.args));
```
sujayVittal, line 24: Same as the comment given in src/wallet/test/init_test_fixture.cpp -> shouldn't it be an address to m_node that needs to be passed to MakeWalletClient(NodeContext, Chain, ArgsManager) method?

sujayVittal (changes requested):
Have some questions posted in - 

- src/wallet/context.h
- src/wallet/test/init_test_fixture.cpp
- src/wallet/test/wallet_test_fixture.h

src/wallet/context.h
rnicoll: Without it I get:

```
In file included from wallet/context.cpp:5:
./wallet/context.h:28:5: error: ‘NodeContext’ does not name a type; did you mean ‘WalletContext’?
   28 |     NodeContext* nodeContext{nullptr};
      |     ^~~~~~~~~~~
      |     WalletContext
```

However I have switched this out for the full include of `node/context.h` and that should be cleaner, thanks for the feedback.

src/wallet/test/init_test_fixture.cpp
rnicoll: So it's passing a reference (`m_node` is already a reference). I could make it into a pointer, but I think this works? Let me know if I'm doing something daft without realising it.

rnicoll (commented):
Addressed initial feedback, please take a look

src/wallet/init.cpp
```diff
@@ -107,7 +107,7 @@ void WalletInit::Construct(NodeContext& node) const
         LogPrintf("Wallet disabled!\n");
         return;
     }
-    auto wallet_client = interfaces::MakeWalletClient(*node.chain, args);
+    auto wallet_client = interfaces::MakeWalletClient(node, *node.chain, args);
```
nooperation, line 110: Will there ever be a situation where you would want to pass a  `chain` that differs from the chain referenced by `node`? If not, then this seems to add a potential place for someone to do something incorrectly, compared to something like `interfaces:MakeWalletClient(node, args);`. 

This would also lightly go for the same for `nodeContext` and `chain` in `WalletContext` - `chain` would ultimately just be a shortcut for nodeContext->chain, but that one would be a much more impactful change

rnicoll: The idea was to build next to the code as it's a Dogecoin change, but replacing also works. Agreed on the shortcut approach being a bigger change, and keeping in mind how we merge from Bitcoin Core is a consideration, one I'd rather avoid.

I've removed chain from being its own parameter until the final `WalletClientImpl` construction, where it was easier to leave it IMHO.

nooperation: I didn't even take into consideration the merging. Whatever keeps that process simpler is probably the best way 👍

langerhans (approved):
ACK with the possible deprecation of getauxblock in mind so we get separation of concerns back later.

src/wallet/test/init_test_fixture.cpp
sujayVittal: I guess this works. We can leave it here for now.

src/wallet/context.h
sujayVittal: Yep, no worries. Looks good for now.

PR #2260, opened 2021-06-06
fix: Boost url dependency

mmicael1:
Address https://github.com/dogecoin/dogecoin/issues/2229 https://github.com/dogecoin/dogecoin/discussions/2256


rnicoll (approved):

rnicoll:
Thanks!

PR #2274, opened 2021-06-08
1.21 default address type

rnicoll:
Switch the default address type for 1.21 from BECH32 (SegWit) to pubkeyhash ("legacy") addresses. This is required as otherwise Dogecoin Core would create addresses by default which cannot be used.

The ability to produce BECH32 addresses is retained for forward compatibility and to reduce scale of changes.

Some RPC tests (especially those using descriptors) require SegWit addresses for them to function as-is, without significant change, and in those cases the address type is set via extra parameters.


langerhans (commented):
Reminder that this needs changing in Qt as well

langerhans (dismissed):

langerhans (approved):
re-approved

soljarjb13 (commented):
Learning as I go with my pocket stylus

PR #2290, opened 2021-06-10
[doc] Doc automation PoC

p-j01:
## Description 

Proof of concept of documentation generation with Doxygen and custom theme.  

## Demo

Available on GitHub Pages of my [fork]( https://p-j01.github.io/dogecoin/).

## Next Steps
Filter current documentation


AbcSxyZ:
Interesting contribution, but it would be great to have a discussion about tools before using Doxygen, to have some input. It seems an interesting/standard tool for C++ docs, so it can end up to use it, but have a comparison of pro/cons of different tools can be great.
I tried to find bitcoin's discuss on that decision, didn't find a thread, does anyone have a link ?

It also forces devs to use a documentation standard within the code, they need to be aware and accept it.

Also, I've in mind a users docs. Possibly it will be separate from the devs docs, but it will be really interesting to have a tool where non-git users can contribute. I've no clue about how to achieve that... a wiki like interfaced with git, form my pov it needs some brainstorming.

PR #2298, opened 2021-06-12
Enforce minor Dogecoin parameters

rnicoll:
* Minimum number of blocks and blockchain size are increased for pruning. In comparison to Bitcoin this uses 24 hours as a minimum, rather than 48, although given blocks are rarely full this likely reflects a lot longer in reality.
* Multiply fork detection parameters by 5. The wall clock time elapsed for "long" forks is therefore half that of Bitcoin, but IMHO those figures are excessive for a chain with 1 minute block times.
* BIP30 has been enabled on Dogecoin since inception and should not be conditional.


rnicoll:
I think the failure is a Travis glitch, works locally, but will leave until #1974 is in and it needs a rebase anyway.

src/validation.h
```diff
@@ -81,18 +81,18 @@ static const bool DEFAULT_FEEFILTER = true;
...
+// Require that user allocate at least 2,200 MiB for block & undo files (blk???.dat and rev???.dat)
+// At 1MB per block, 1,440 blocks = 1,440MB.
+// Add 15% for Undo data = 1,656MB
+// Add 20% for Orphan block rate = 1,987MB
+// We want the low water mark after pruning to be at least 1,987 MB and since we prune in
 // full block file chunks, we need the high water mark which triggers the prune to be
-// one 128MB block file + added 15% undo data = 147MB greater for a total of 545MB
-// Setting the target to >= 550 MiB will make it likely we can respect the target.
-static const uint64_t MIN_DISK_SPACE_FOR_BLOCK_FILES = 550 * 1024 * 1024;
+// one 128MB block file + added 15% undo data = 2,134MB greater for a total of 545MB
```
nooperation, line 93: `2,134MB greater for a total of 545MB`

I think the wrong value was changed in the comment here. One 128MB block file + its 15% undo data will still be 147MB, but the total of 545MB should be changed to a total of 2,134MB? Or would it be 2,135MB because the original comment added 1MB, maybe to account for the previous rounding?

test/functional/p2p_node_network_limited.py
```diff
@@ -4,9 +4,9 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Tests NODE_NETWORK_LIMITED.
 
-Tests that a node configured with -prune=550 signals NODE_NETWORK_LIMITED correctly
+Tests that a node configured with -prune=2250 signals NODE_NETWORK_LIMITED correctly
```
nooperation, line 7: Should this be testing -prune=2200 instead of -prune=2250 to be more consistent with the original test? (also down below on line 38)

rnicoll: I think you're right, I'll dig into that.

PR #2328, opened 2021-06-21
Docs4Doge : Update unix build instructions

AbcSxyZ:
Starting to dive into installation instructions for [docs4doge](https://github.com/dogecoin/dogecoin/issues/2271), focus only on unix.

The goal is to reduce entry barrier for users, make it easier to getting started with Dogecoin core.

I reorganized the content as follows to be more logical : *Dependencies install > Build Steps > build Configuration*.

## Dependencies

I made some choice for dependencies, need feedback/check : 
- Use only qt5 and remove qt4 (#2316)
- For BerkeleyDB, recommend 5.3 with package manager (5.1 not available [on debian/ubuntu/[gentoo/fedora](https://github.com/dogecoin/dogecoin/pull/2221#issuecomment-851028567)]), but suggest 5.1 for source install. (See #2245)
- For boost, install package separately only (`libboost-system-dev libboost-filesystem-dev [...] libboost-thread-dev`) instead of `libboost-all-dev`

## Removed notes
I removed a couple of notes, like the first one who explain that you must use a right path.
Also I removed `The release is built with GCC and then "strip dogecoind" to strip the debug symbols, which reduces the executable size by about 90%.`, I thought it wasn't an important info for compilation, idk if we should keep it.

PS. : ARM compilation is not at the right place in my mind, at the end because still WIP.
**[Tested only with debian]**

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -1,35 +1,22 @@
-UNIX BUILD NOTES
-====================
-Some notes on how to build Dogecoin Core in Unix.
+# Unix Build Notes
+1. [Install dependencies](#dependencies)
+ - [Ubuntu & Debian](#ubuntu)
+ - [Fedora](#Fedora)
+ - [FreeBSD](#FreeBSD)
+2. [Build steps](#build-steps)
+3. [Build configuration](#configure-build)
+ - [Enable Qt Gui](#enable-qt-gui)
+ - [Incompatible BerkeleyDB](#incompatible-berkleydb)
+ - [Disable Wallet](#diable-wallet)
+ - [Miniupnpc](#miniupnpc)
+ - [Security](#security)
 
-(for OpenBSD specific instructions, see [build-openbsd.md](build-openbsd.md))
+## Install dependencies
 
-Note
----------------------
-Always use absolute paths to configure and compile Dogecoin and the dependencies,
-for example, when specifying the path of the dependency:
+You must install required dependencies to run a basic Dogecoin daemon.  
+Optional dependencies may vary depending on your build specifications.
 
-	../dist/configure --enable-cxx --disable-shared --with-pic --prefix=$BDB_PREFIX
-
-Here BDB_PREFIX must be an absolute path - it is defined using $(pwd) which ensures
-the usage of the absolute path.
-
-To Build
----------------------
-
-```bash
-./autogen.sh
-./configure
-make
-make install # optional
-```
-
-This will build Dogecoin-Qt as well if the dependencies are met.
-
-Dependencies
----------------------
-
-These dependencies are required:
+**Required dependencies :**
 
  Library     | Purpose          | Description
  ------------|------------------|----------------------
@@ -37,7 +24,7 @@ These dependencies are required:
  libboost    | Utility          | Library for threading, data structures, etc
  libevent    | Networking       | OS independent asynchronous networking
 
-Optional dependencies:
+**Optional dependencies:**
 
  Library     | Purpose          | Description
  ------------|------------------|----------------------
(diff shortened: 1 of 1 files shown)


alamshafil:
I think you should quickly talk about how people can use the `depends` system to build static version of Dogecoin. There is a README.md in the `depends` where you can link people to. This is an idea, it might not be needed inside the UNIX build doc.

doc/build-unix.md
```diff
@@ -108,27 +101,58 @@ According to installed dependencies, the following steps will compile `dogecoind
...
+```bash
+# Install script for BerkeleyDB 5.1
+
+# BerkeleyDB installation directory
+BDB_PREFIX=$(pwd)/bdb
+mkdir $BDB_PREFIX
+
+# Fetch the source and verify shasum
+wget 'http://download.oracle.com/berkeley-db/db-5.1.29.NC.tar.gz'
+echo '08238e59736d1aacdd47cfb8e68684c695516c37f4fbe1b8267dde58dc3a576c db-5.1.29.NC.tar.gz' | sha256sum -c
```
fmhc, line 132: this is working
image

fmhc (commented):
For me this looks good

fmhc:
> I think you should quickly talk about how people can use the `depends` system to build static version of Dogecoin. There is a README.md in the `depends` where you can link people to. This is an idea, it might not be needed inside the UNIX build doc.

if you want to add that, you could also add a sentence to further information on testing, as this could be something people want to do after compiling for unix

AbcSxyZ:
It's not entirely clear what `depends` folder used for, but it does not seem restricted to UNIX but are cross-platform information concerning installation. Actually, it's possible that build steps (`autogen/configure/make`) and this kind of information should be exact from UNIX build instructions, to avoid duplicate information and be DRY

alamshafil:
The depends system allows users to build or cross-build a static version of Dogecoin that will work on any computer. If you compile without depends, then the dogecoin binaries you build might not work on another computers. You can say if you want to compile a binary that will work anywhere, use depends (then link to `README.md`). The depends can be used to easily to cross-compile static binaries for any OS or computer.

chromatic:
@AbcSxyZ Looks like this PR is a little bit out of date with regard to the current state of the branch. Would you mind rebasing?

AbcSxyZ:
Who cares ? I'm unmotivated to do it because I feel like (some) external contributor work is not really taken in consideration by maintainers (I've also #2396 in mind to not speak only about my work). No idea of the reasons (time, priority ?), but I'm not comfortable with the situation. It seems I'm not the lonely one who run away (who try to contribute since 1 or 2 months ?).

So right now, I don't mind 🙃. I will let them doing it when and if they want to integrate it. It may be outdated in a couple of months again 😉 !

It takes effort and it's a work on his own, but open source projects should empower contributors :)

chromatic:
@AbcSxyZ I totally get that. I intended to provide feedback on this and your other documentation PRs, but life has been busy and I apologize.

I think this PR has a lot of value, and I'd be happy to cherry-pick it into a new PR so that the work gets done and you still get the credit -- just me taking on the administrative work to get it merged. Would that work for you? I'm happy to do that on the other PRs as well.

patricklodder:
> I'm unmotivated to do it because I feel like (some) external contributor work is not really taken in consideration by maintainers

Hi! I'm sorry about that, is not intended as such. This PR (and your README update too) is on my todo list to get done for 1.14.4 still, but it is not very high on the list while I still have to deal with bugs, reversal of the 1.14.2 fee fiasco and release quality assurance. I really do want to get this in; as you may have noticed, I have descoped a lot of PRs from 1.14.4 because they are either not ready or not important enough, but I have explicitly not descoped this. It means that I want to get to it. I currently prioritize `bugs > qa > quick wins > enhancements > docs > contrib`, so this PR and the other you mention are on the bottom of the list. 

Also note that this is not a quick win because these install instructions need to be tested - even if they didn't change, doesn't mean they work because quality of documentation has been even worse than code quality for v1.14.x. That means that since no one has reviewed and verified the actual instructions, this now sits on my plate to do if I want to move this forward. The Debian/ubuntu is easy for me, but I now have to find myself a way to get a freebsd and a fedora install - neither of which I have worked on for years - and make sure that this works. I intend to do this, I hope I can make it - bugs and quality control are dominating my schedule at the moment.

ReverseControl:
@patricklodder Now that we have addressed bugs, qa, could you revisit this? Where are we?

@michilumin @rnicoll

patricklodder:
> Now that we have addressed bugs, qa, could you revisit this? Where are we?

Been working on this all day.

patricklodder:
Just to keep everyone informed of progress, here's where I'm at with this:

1. The instructions for FreeBSD prerequisite packages have been overwritten with those for Fedora; wrong package manager, package names...
2. The Fedora bdb packages do not reflect Dogecoin (not in the original document either) AND no longer exist - I know how to fix these and have verified those to work - for building without GUI. I have not yet had time to verify QT builds on Fedora.
3. `./autogen.sh` does not work anymore on Fedora 34 because `glibtoolize` is now called `libtoolize`. However, running `autoreconf --install --force --warnings=all` manually does work.
4. The original FreeBSD instructions do not work for wallet compilation either under 11.4 (original doc says it works for 11.0, but that's Bitcoin's statement, not Dogecoin's, except for someone doing `s/Bitcoin/Dogecoin/g`) - I do not yet have a fully working fix and my `qemu` image is terribly slow so this is eating away time. I am currently building with `--disable-wallet` to make sure at least that part works and have to focus on building with the native `db5` package later.

Even though I'd love to, I don't think we can unify and simplify the build process in the way it's proposed here because it really is different for each unix flavor / distro. Can we talk alternatives? Maybe we can build these docs on a per-OS basis outside of the release process (wiki, maybe?) and then link that in the README.md? Open for suggestions.

chromatic:
:+1: to moving specific OS instructions elsewhere.

The drawback of a wiki is that it's vulnerable to bad actors giving bad advice in a way that officially released tarballs isn't -- but if we encourage people to use official packages instead, then hopefully only developers will find the specific instructions.

Perhaps listing dependencies in general (Boost, Qt, OpenSSL, bdb, etc) is sufficient, with links to find more platform-specific information. That pushes platform-specific details to a place where they aren't a process bottleneck in an ever-growing document.

The rest of the configuration/compilation steps can remain here.

ReverseControl:
@patricklodder I agree with moving this work out of the releases, this one and all of them, and have a parallel project dedicated to the documentation where we would merge as go along into the releases. It is counter-productive to let the perfect be the enemy of all the hard work we have done to delay releases when our community is eagerly waiting for a better system.

patricklodder:
See #2460

AbcSxyZ:
@patricklodder About "empowering external contributor", I will create a discussion later, I want to do a couple of feedback in the hope to improve the situation of contributors.

My bad for switching FreeBSD and Fedora instructions, fixed.
Idk how the unification is possible, I don't use those distros, I was focused on refreshing Debian & derivated distros, and kept previous instructions for FreeBSD & Fedora.

About splitting instruction, idk how big would be the difference. I think at first we can think about potential wins for user. In my mind, we should rework the `INSTALL.md`, to have instruction about downloading the core from release, then build from sources for main distros. If we have clear instruction for Windows/Debian related users, I guess we cover more than 95% of need. So for a while, regarding the rest of the doc which also need a lot of work, I guess keeping outdated Fedora/FreeBSD instructions outdated for is not a big deal (flag it with  WIP/issue, maybe a Fedora/FreeBSD user would update it, it may show if there is a need also).

So for the split of instruction by OS, if there is big difference, maybe in the future it would be interesting, for now maybe not. Depend on if you really want to have updated Fedora/FreeBSD instruction rn.
The current PR keep the previous unification.

patricklodder:
I've unblocked as much as a can from this PR with #2465. Hope that's acceptable.

patricklodder:
Merged with #2465 - thank you for the work! Much appreciated!

Issue #2338, opened 2021-06-24
[bug] [1.14.4-dev] Qt keyboard detection issue /w german keyboard layout


fmhc:
# Bug Report

**Current behavior:**
When installing the 1.14.4-dev on Ubuntu 20.04 with german keyboard layout, the dogecoin-qt starts / loads and syncs correctly but throws an error on the console and the keyboard input can not be used.

**Expected behavior**
The keyboard layout is correctly loaded from Qt and the user is able to use the keyboard to type into form fields and use buttons.

**Steps to reproduce:**
1. Install fresh Ubuntu 20.04 LTS Desktop VM
2. Set the keyboard layout / language to German (standard) while setup
3. snap install dogecoin-unofficial --channel=edge
4. dogecoin-unofficial.qt
5. open a form (e.g. send) and try typing in an address with the keyboard = not working

**Screenshots**
![image](https://user-images.githubusercontent.com/3387385/123192893-152e1a00-d4a4-11eb-8957-9cf806265b33.png)

**Dogecoin Core version**
Version 1.14.4-dev PR 1904: https://github.com/dogecoin/dogecoin/pull/1904
probably same results with all 1.14.4-dev as all based on Qt v5.7.1

**Machine specs**
- OS: Ubuntu 20.04 LTS Desktop 
- CPU: 4 cores VM
- RAM: 4 GB
- Disk size: 100 GB
- Disk Type (HD/SDD): HDD VMI

**Extra information**
related issue in Qt v.5.7.1 (we're using that in 1.14.4-dev): https://github.com/conda-forge/qt-feedstock/issues/104
we need to upgrade to a newer version of Qt. See discussion in https://github.com/dogecoin/dogecoin/pull/1904 also

Newer version of Qt shoudl be changed here: https://github.com/dogecoin/dogecoin/blob/1.14.4-dev/depends/packages/qt.mk
will try that out with qt 5.15 or 5.12, I'm currently confused about the structure of the newer versions on the download archive server...

fmhc:
https://github.com/katzenmalen/dogecoin/commit/9d138a75a2078532cdde52ef913c95490215eb56
![image](https://user-images.githubusercontent.com/3387385/123241812-c010f900-d4e1-11eb-9af2-1242b244ada4.png)
waiting for the rpc tests but i guess that should be fine. Tested with building on Ubuntu 20.04. / german keyboard layout and locale.

[...]

PR #2339, opened 2021-06-25
upgraded qt to v5.15.2

fmhc:
https://github.com/dogecoin/dogecoin/issues/2338
Unit tests successful
image
image
image
RPC tests successful
image

working dogecoin-qt

image

and typing with the Keyboard works again:
image

### DIFF
--- a/depends/packages/qt.mk
+++ b/depends/packages/qt.mk
@@ -1,6 +1,6 @@
 PACKAGE=qt
-$(package)_version=5.12.11
-$(package)_download_path=http://download.qt.io/new_archive/qt/5.12/$($(package)_version)/submodules
+$(package)_version=5.15.2
+$(package)_download_path=http://download.qt.io/new_archive/qt/5.15/$($(package)_version)/submodules
 $(package)_suffix=everywhere-src-$($(package)_version).tar.gz
 $(package)_file_name=qtbase-$($(package)_suffix)
 $(package)_sha256_hash=909fad2591ee367993a75d7e2ea50ad4db332f05e1c38dd7a5a274e156a4e0f8
--- a/depends/packages/qt.mk
+++ b/depends/packages/qt.mk
@@ -1,21 +1,22 @@
 PACKAGE=qt
-$(package)_version=5.15.2
-$(package)_download_path=http://download.qt.io/new_archive/qt/5.15/$($(package)_version)/submodules
-$(package)_suffix=everywhere-src-$($(package)_version).tar.gz
+$(package)_version=5.12.11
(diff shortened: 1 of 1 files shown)


AbcSxyZ (commented):
Nice work, change looks good, I checked the url change. Hash are also correct, I confirm the following : 
```
0eb522ff6c2194f9690cbdcd5b33e803  qtbase-everywhere-src-5.15.2.tar.xz
739403634e1326a3d52902a835015a4b  qttools-everywhere-src-5.15.2.tar.xz
9b66cdb64402e8fd9e843f8a7120abb1  qttranslations-everywhere-src-5.15.2.tar.xz
```

I'm learning about how to use `depends` folder, wasn't able to test compilation right now. Will try in the following days to run unit tests on my side.

patricklodder:
Note: this needs gitian checks. Adding do not merge label for now.

fmhc:
sorry i guess i commited the wrong version here for v5.12.11 and not v5.15.2:

image

fmhc:
This leads me to the question why i could compile it while the sums were wrong

rnicoll:
I'm noting Bitcoin hasn't gone past `5.12.11` (https://github.com/bitcoin/bitcoin/blob/master/depends/packages/qt.mk) and I wonder if they know something. In brighter news, I can see a few broken download URLs but they're easier to fix.

rnicoll:
The answer to why it worked locally is that it _normally_ uses the locally available version of Qt, it's only for CI, Gitian and a few other cases that it uses specific fixed versions.

rnicoll:
> I'm noting Bitcoin hasn't gone past `5.12.11` (https://github.com/bitcoin/bitcoin/blob/master/depends/packages/qt.mk) and I wonder if they know something. In brighter news, I can see a few broken download URLs but they're easier to fix.

Looks like it was just a timing issue: https://github.com/bitcoin/bitcoin/pull/22054

rnicoll:
I had a go at fixing the CI breakage, but we're going through so many versions here it's hard to work out what patches we'll need. Potentially this is too ambitious a jump to do all at once. I've put my work in progress into a PR in case it helps, but don't anticipate completing it myself: https://github.com/katzenmalen/dogecoin/pull/1

micaelmalta:
The upgrade is not so simple: See https://github.com/bitcoin/bitcoin/pull/21376/files

patricklodder:
removed from 1.14.4 target milestone, qt (and openssl) dependency updates will have to target 1.14.5

patricklodder:
Closing this for now. Foreign keyboard support has been fixed with 1.14.5's libxkb inclusion, and 5.15 is unmaintained

PR #2401, opened 2021-07-14
Refine Snap verification message

rnicoll:
Refine Snap verification message to accurately reflect it's checking secure hashes,
not signatures. Checking signature would entail verifying there is a signature from a
trusted person, not just that the hash matches a value in the known values list.


rnicoll:
Follows on from #1904

patricklodder (approved):
utACK

PR #2422, opened 2021-07-24
test: Fixup replace-by-fee values

rnicoll:


rnicoll:
Opening early while I work out the remaining test case.

patricklodder (commented):
Can we move this from extended tests to standard tests so we get CI coverage? Otherwise ack (tested locally on Ubuntu 20.04)

rnicoll:
@patricklodder Done

patricklodder (approved):
ACK. tested on Ubuntu 20.04.

PR #2433, opened 2021-07-31
1.14.4 mempool packages

xanimo:
i apologize this is not atomic, i can redo .gitignore in separate pr if need be, but if not i noticed after setting up qa dependencies, vs code was tracking files within `qa/pull-tester/ltc-scrypt-master/` and after forgetting to omit benchmark in `./configure`, vs code was also tracking `src/bench` binary so added those to .gitignore.

as for the mempool_packages.py test, i bumped the fee from .0001 to 1 to fix insufficient priority err.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,7 +102,6 @@ linux-coverage-build
 linux-build
 win32-build
 qa/pull-tester/tests_config.py
-qa/pull-tester/ltc-scrypt-master/
 qa/cache/*
 
 !src/leveldb*/Makefile
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ src/dogecoin
 src/dogecoind
 src/dogecoin-cli
 src/dogecoin-tx
+src/bench
 src/test/test_dogecoin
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -5,6 +5,7 @@ src/dogecoin
 src/dogecoind
 src/dogecoin-cli
 src/dogecoin-tx
+src/bench
```
rnicoll, line 8:
```suggestion
```
Lets cut these back out, and we can the squash the lot in GitHub and commit them later.

.gitignore
```diff
@@ -102,6 +103,7 @@ linux-coverage-build
 linux-build
 win32-build
 qa/pull-tester/tests_config.py
+qa/pull-tester/ltc-scrypt-master/
```
rnicoll, line 106:
```suggestion
```

rnicoll (changes requested):
I think good if we revert the `.gitignore`, simply because otherwise they're 2:1 unrelated lines to related lines changed. Then good to go I think.

rnicoll (approved):

PR #2482, opened 2021-08-21
 [ Security ] Increased entropy from OpenSSL and OS to a full SHA512 …

ReverseControl:
…block (64 bytes) and added additional entropy from RNSEED on CPUs that support it.

To compile it a flag is needed to enable the RNSEED instruction. I compiled with 

```
./configure CFLAGS='-mrdseed ' CXXFLAGS='-mrdseed ' CPPFLAGS='-mrdseed' --with-incompatible-bdb
```

As for the actual changes:

1. It is unclear to me why we only used half a SHA512 block to gather entropy from OpenSSL, and the same for OS entropy.
2. I increased that entropy gathering to a full block for each one.
3. I added entropy from the RDSEED instruction implemented on x86 by Intel and AMD.
4. The entropy from RDSEED is NIST compliant. ( See https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/details?product=11882)
5. All AMD Zen processors support this instruction and most Intel Processors launched since 2015. 
6. The existence of this instruction is checked at runtime, if it is not available we default to the previous mechanism where we gather entropy from OpenSSL and the OS. Only we gather twice as much as before to get one full SHA512 block when hashing.

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -23,6 +23,7 @@
 
 #include <openssl/err.h>
 #include <openssl/rand.h>
+#include <lowlevel.h>
 
 static void RandFailure()
 {
@@ -129,20 +130,75 @@ void GetRandBytes(unsigned char* buf, int num)
     }
 }
 
+
+/* /fn int getRNSEED( uint64_t *buff, int num)
+ * /brief Use Intel's SP 800-90B & C compliant Hardware implemented
+ *        instruction to add entropy to the pool use to create keys. 
+ *
+ *        For details, see: https://software.intel.com/content/www/us/en/develop/blogs/the-difference-between-rdrand-and-rdseed.html
+ * 
+ *  /param buff Buffer to fill with random bits.
+ *  /param num  Number of 64-bit words to write to buffer.
+ *  /return     Returns 1 if the hardware returned  a valid stream of random numbers, 0 otherwise.
+ *
+ */
+int getRDSEED( uint64_t *buff, uint32_t num){
+    //Check to see if the processor running this code has 
+    //this Instruction. Most Intel/AMD CPUs after 2015
+    //should have this instruction.
+    uint32_t level, eax, ebx, ecx, edx; 
+    
+    //Level where the RNSEED feature is described
+    level = 1;
+
+    //Extract feature vector
+    __get_cpuid(level, &eax, &ebx, &ecx, &edx);    
+
+    //Initialize state to 0, or failed.
+    uint32_t  cumulativeStatus = 0;
+
+    //Check if RNSEED exists on this processor
+    if( (ecx & bit_RDSEED) == bit_RDSEED ) {
+
+        //Re-initialize value to 1, or success 
+        cumulativeStatus = 1;
+
+        for( uint32_t kk = 0; kk < num; kk++){
+            //Get rng data
+            unsigned long long int rng;
+            uint32_t status = _rdseed64_step( &rng );
+
+            //Store seed data
+            buff[kk] = rng; 
+
+            //Update status value
+            cumulativeStatus &= ( status == 1) ? 1 : 0;
+        }
+    } 
+
+    return cumulativeStatus;
+}
+
 void GetStrongRandBytes(unsigned char* out, int num)
 {
     assert(num <= 32);
     CSHA512 hasher;
+    block_512bits buf2;
     unsigned char buf[64];
 
     // First source: OpenSSL's RNG
     RandAddSeedPerfmon();
-    GetRandBytes(buf, 32);
-    hasher.Write(buf, 32);
+    GetRandBytes(buf, 64);
+    hasher.Write(buf, 64);
 
     // Second source: OS RNG
     GetOSRand(buf);
-    hasher.Write(buf, 32);
+    hasher.Write(buf, 64);
+
+    //Third Source: Intel's NIST SP 800-90B & C compliant RBG
+    //if available, use it.
+    if ( getRDSEED(buf2.qword, 64) == 1  )
+        hasher.Write(buf2.ubyte, 64);
 
     // Produce output
     hasher.Finalize(buf);


ReverseControl:
I am looking at all the errors...its a different error for each system. hahahhahahahahaha


Okay,  here we go:

1. The ARM errors: I think the thing to do here is to do compile time headers and default to the previous version with double the entropy gathering (See notes on the PR).
2. For the 32-bit systems, I will default back to 32-bit registers to gather the data.
3. For 64-bit systems, the issue is the compiler flag I mentioned in the PR.
4. Apparently MacOS does not have the same cpuid.h as found in linux, which makes sense. In particular, with the new processors like the M1, I think the best course of action is the same one as for ARM. Default back to previous method with double the entropy gathering.
5. The 64-bit windows error is the same as the one I got. Solving the compiler flag will solve this one too.


@patricklodder @michilumin @edtubbs

ReverseControl:
We are now getting the same error on all but the MacOS platform. Those errors will be fixed when the flags are added to the compiler.

As for MacOS, the fix is defining the ``bit_RDSEED`` macro as it seems ``#include <cpuid.h>`` on MacOS is not defined. I will just define it myself if it is not defined.

ReverseControl:
@patricklodder @rnicoll Okay, ARM and MacOS are ready to go. We need to add the compiler flags to fix the others. Alternatively, we could just use the byte code implementation of RDSEED, but I'd rather fix the intrinsic flag issue now as it will come up again.

patricklodder:
like with sse2, I'd like a ./configure flag for this
unlike with sse2, we should probably give some thoughts about what the default can/should be

ReverseControl:
@patricklodder Given this particular implementation checks for availability of this instruction at runtime, the default on the compiling side should be to always enable it. How that is done is up to discussion.

patricklodder:
I think I can be okay with a default enable, as long as it also can be switched off, to enhance portability to architectures we currently do not test for / do not foresee. So for that to be nice, we'd just add `--disable-rnseed` to configure.ac, which can override a use_rnseed=yes as a configure flag. Kinda like how zmq integration is done - default on, can be disabled.

Thoughts?

ReverseControl:
@patricklodder  That's perfect.

ReverseControl:
@patricklodder @rnicoll  How do I detect ARM at the configure.ac level? That is the only way I can think of to solve this. I can think of another way, but it would probably break the Windows build.

configure.ac
```diff
@@ -859,6 +866,19 @@ else
   fi
 fi
 
+
+
+dnl Use rdseed to gather entropy
+if test "x$use_rdseed" = "xyes"; then
+        ARM_CC=`which arm-linux-gnueabihf-g++`
```
patricklodder, line 873: I think this is incorrect. The fact that I have an ARM compiler installed does not mean I am compiling for ARM. What was not functioning with $host detection?

src/lowlevel.h
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) 2021 The Dogecoin Core developers
```
patricklodder, line 1: Right now, the functionality defined by this header file is only used in `random.cpp`, what is the benefit of creating a new header file over using `random.h`?

src/lowlevel.h
```diff
@@ -0,0 +1,55 @@
...
+        /* To address compatibility issues with MacOS, given the lacking implementation
+         * of intrinsics we here implement rdseed in byte code so that we may use it 
+         * anywhere the ISA extension permits it.
+         *
+         * See: https://software.intel.com/content/www/us/en/develop/articles/the-drng-library-and-manual.html
+         * You will find this function definition at:
+         *    libdrng-1.0.tar.gz: https://software.intel.com/file/469237/download
+         */
+        #  define __rdseed16_step(x) ({ unsigned char err; asm volatile(".byte 0x66; .byte 0x0f; .byte 0xc7; .byte 0xf8; setc %1":"=a"(*x), "=qm"(err)); err; })
+        #  define __rdseed32_step(x) ({ unsigned char err; asm volatile(".byte 0x0f; .byte 0xc7; .byte 0xf8; setc %1":"=a"(*x), "=qm"(err)); err; })
```
patricklodder, line 40: Question: why not define this inside a `#if defined(__APPLE__)` block, and wrap the `_rdseed*_step(x)` inside this same `__rdseed*_step(x)`, so that the implementation in the cpp file becomes much cleaner?

src/lowlevel.h
```diff
@@ -0,0 +1,55 @@
...
+//mana potions and doggy treats...especially magical doggy treats.
+
+
+//Only define this header on x86
+#if ( defined(i386)     || defined(__i386)  || defined(__i386__) || \
+      defined( __386)   || defined(_X86_)   || defined( _M_I86)  || \
+      defined(__i386__) || defined(__X86__) || defined(__x86_64) ) 
+
+    #ifndef LOWLEVEL_H
+        #define LOWLEVEL_H
```
patricklodder, line 16: This header guard currently also serves as a functionality flag in `random.cpp`, recommend to define a separate functionality flag, eg `#define USE_RDSEED`, right underneath this, as that would keep flags serving a single purpose.

src/random.cpp
```diff
@@ -129,22 +130,95 @@ void GetRandBytes(unsigned char* buf, int num)
...
+ *  /param buff Buffer to fill with random bits.
+ *  /param num  Number of 64-bit words to write to buffer.
+ *  /return     Returns 1 if the hardware returned  a valid stream of random numbers, 0 otherwise.
+ *
+ */
+int getRDSEED( uint64_t *buff, uint32_t num){
+    //Initialize state to 0, or failed.
+    uint32_t  cumulativeStatus = 0;
+
+    #ifdef LOWLEVEL_H
```
patricklodder, line 149: should be changed to a function flag rather than using the header guard.

src/random.cpp
```diff
@@ -129,22 +130,95 @@ void GetRandBytes(unsigned char* buf, int num)
...
+    //Third Source: Intel's NIST SP 800-90B & C compliant RBG
+    //if available, use it.
+    #ifdef LOWLEVEL_H    
+        block_512bits buf2;
+        if ( getRDSEED(buf2.qword, 8) == 1  )
+            hasher.Write(buf2.ubyte, 64);
+    #endif
 
-    // Produce output
+    // Produce output
```
patricklodder, line 221: Unneccessary whitespace change

src/random.cpp
```diff
@@ -129,22 +130,95 @@ void GetRandBytes(unsigned char* buf, int num)
...
+    hasher.Write(buf, 64);
 
     // Second source: OS RNG
     GetOSRand(buf);
-    hasher.Write(buf, 32);
+    hasher.Write(buf, 64);
+
+    //Third Source: Intel's NIST SP 800-90B & C compliant RBG
+    //if available, use it.
+    #ifdef LOWLEVEL_H
```
patricklodder, line 215: should be changed to a function flag rather than using the header guard.

patricklodder (changes requested):
Code review only so far, have not yet tested this in-depth.

configure.ac
ReverseControl: I tried several versions of the ``uname`` command which tells architecture, but none of them worked. Since I don't have a ARM based system, I rented an AWS ARM Linux machine and this was the only trial that worked both on that one and on the CI pipeline here. As for the ``$host`` variable, it is not defined. I am not quite sure what you mean by it.

src/lowlevel.h
ReverseControl: I expect changes that I will be making in the future to use low level functionality including intrinsics, assembly and other fancy things. I'd rather have one place for low level stuff, it just so happens that first place where I implemented such functionalities is related to random.h(.cpp).

src/lowlevel.h
ReverseControl: That is a good suggestion. I will do that. I don't think we will use this particular instruction anywhere else in this form, so that should be fine.

src/random.cpp
ReverseControl: Narrowing scope is better. I will do that.

src/random.cpp
ReverseControl: There is a struct here that is part of the header file, and not a particular function or soon to defined function guard for rdseed.

src/lowlevel.h
ReverseControl: Good suggestion. I will do that.

configure.ac
ReverseControl: @patricklodder Any suggestions on how to make this work would be greatly appreciated.

src/random.cpp
```diff
@@ -129,20 +130,87 @@ void GetRandBytes(unsigned char* buf, int num)
...
+                int32_t status1 = _rdseed32_step( &rng_low );
+                int32_t status2 = _rdseed32_step( &rng_high );
+
+                //Place values into a 64-bit register
+                uint64_t temp = rng_high;
+                temp <<= 32;
+                temp += rng_low;
+
+                //Store into buffer
+                buff[kk] =  temp;
```
RichDevX, line 180:
```suggestion
                buff[kk] =  ((uint64_t)rng_high << 32) | rng_low;

```

src/random.cpp
```diff
@@ -129,20 +130,87 @@ void GetRandBytes(unsigned char* buf, int num)
...
+    
+        //Check if RDSEED exists on this processor
+        if( (ecx & bit_RDSEED) == bit_RDSEED ) {
+    
+            //Re-initialize value to 1, or success 
+            cumulativeStatus = 1;
+    
+            for( uint32_t kk = 0; kk < num; kk++){
+                //Get RBG data
+                unsigned int rng_high, rng_low;
```
RichDevX, line 169:
```suggestion
                uint32_t rng_high, rng_low;
```

src/random.cpp
```diff
@@ -129,20 +130,87 @@ void GetRandBytes(unsigned char* buf, int num)
...
+        __get_cpuid(level, &eax, &ebx, &ecx, &edx);    
+    
+        //Check if RDSEED exists on this processor
+        if( (ecx & bit_RDSEED) == bit_RDSEED ) {
+    
+            //Re-initialize value to 1, or success 
+            cumulativeStatus = 1;
+    
+            for( uint32_t kk = 0; kk < num; kk++){
+                //Get RBG data
```
RichDevX, line 168:
```suggestion
                //Get RNG data
```

src/random.cpp
```diff
@@ -129,20 +130,87 @@ void GetRandBytes(unsigned char* buf, int num)
...
+                unsigned int rng_high, rng_low;
+
+                int32_t status1 = _rdseed32_step( &rng_low );
+                int32_t status2 = _rdseed32_step( &rng_high );
```
RichDevX, line 172: Should place in a retry loop
https://software.intel.com/content/www/us/en/develop/articles/intel-digital-random-number-generator-drng-software-implementation-guide.html#inpage-nav-5-9

Mahandrafhwn:
Unsubscribe

src/random.cpp
ReverseControl: @RichDevX This one is a matter of style. I would normally write as you have it written, but looking at the source code I found very few places where type casting is done, so I wrote it the way I did.

src/random.cpp
ReverseControl: RDSEED is a random bit generator, not a random number generator, per Intel documentation.

src/random.cpp
ReverseControl: The entire function returns a 1 or 0 on success or failure, respectively, it is the caller's responsibility to address that properly. In our case, we have. Look at the calling code, if it fails, we move on without using it.

src/random.cpp
ReverseControl: This one is a matter of style, once again. I do like uintX_t better, so I will change this one.

ReverseControl: @RichDevX I do want to note that on processors where unsigned int has a smaller bit width than 32-bits, this code never executes because the flags of cpuid instruction corresponding to this feature evaluates to zero.

src/random.cpp
RichDevX: The processor can request bits faster than the entropy source can output bits, so it would be ideal to retry until new bits are available instead of discarding all the bits due to one or more failures.

ReverseControl: The question becomes how often should we try before we give up, how many cycles should we wait to retry again, in short how much time is it okay to wait until we get the requested random bits? As it is, if it is not ready when we ask, we move on. I think spending more than 1,000 clock cycles on this is too much. We could run the PAUSE  and WAIT instruction and then try again, though the behavior of the PAUSE instruction is very processor specific in terms of how many cycles it delays instruction execution, but as it is I think it is good enough for now. Plus, cpuid is called before this, so between calls of our function  there is a very long delay in cycles while cpuid clears the execution pipeline, stalls the reorder buffer and waits for memory operations instructions to be retired. We only ask for 512-bits of entropy, and this function does not get called with high frequency. Who generates thousands of wallets a second anyway?

ReverseControl:
@patricklodder Could you confirm: what does ``uname -p`` return on the machine that is running the arm linux test?

src/random.cpp
```diff
@@ -129,20 +130,87 @@ void GetRandBytes(unsigned char* buf, int num)
...
+                uint64_t temp = rng_high;
+                temp <<= 32;
+                temp += rng_low;
+
+                //Store into buffer
+                buff[kk] =  temp; 
+    
+                //Update status value. 
+                //cumulativeStatus will be 1 iff all iterations are 1.
+                cumulativeStatus &= status1 & status2;
```
edtubbs, line 184: should be using the 64-bit instruction...  but, it's also worth mentioning there's an associated CVE (CVE-2020-0543).
int _rdseed64_step (unsigned __int64 * val)

configure.ac
patricklodder: Fixed it for you in https://github.com/ReverseControl/dogecoin-1/pull/15 using `$host`

src/lowlevel.h
patricklodder: I've reopened this comment. The new code can still be optimized but we can do that separately.

src/random.cpp
patricklodder: I have an alternative for this I think but I don't want to stall out any testing on this. Currently it works, we can improve maintainability in a later PR.

src/random.cpp
ReverseControl: The reason for the 32-bit version is that we test on 32-bit systems, and Intel uses it like this in the implementation I reference in the documentation. If you look at the header file for the 64-bit version, they actually call the 32-bit version twice just like this.

configure.ac
ReverseControl: Thanks. Merged.

src/random.cpp
ReverseControl: As for that CVE, the microcode update to deal with that CVE serializes instruction execution to avoid that side channel attack. We execute CPUID right before getting this data, which serializes and flushes out the entire pipeline (memory ops, compute ops, and everything in the reorder buffer) so that will help quite bit to mitigate this.

edtubbs: I suppose it makes sense if you're planning on recommending Intel users perform microcode updates with 14.5.  Respectfully, there's no documented evidence CPUID mitigates this issue at all.

patricklodder: > I suppose it makes sense if you're planning on recommending Intel users perform microcode updates with 14.5. 

Keeping in mind that at best this entropy is only 1/3rd of the total used for wallet creation, would you recommend to:

1. Keep it on for x86 and mention it in the release notes, or
2. Disable it from release and mention it in the build docs?

ReverseControl: @edtubbs Given the paper of the attack itself, which gave rise to this CVE (https://www.vusec.net/projects/crosstalk/), and the fact that they use Transient Execution to read the Line Fill Buffers(LFB), specifically (Rogue In-Flight Data Load, RIDL) see https://github.com/vusec/ridl/blob/bdcc1a78cd8246719282befa7d3af8e7a2f1cb89/pocs/rdrand/leak_code.h. Executing CPUID between RDSEED executions would dramatically decrease the likelihood of data leakage.

So, to be fair, the execution of it only at the beginning is in fact not helpful for mitigation of this CVE.

patricklodder: > So, to be fair, the execution of it only at the beginning is in fact not helpful for mitigation of this CVE.

That CVE is externally mitigated by Intel's patch though, so while I played with the thought of asking "how can we force a flush directly AFTER the call to RDSEED", I don't think that's really needed. Workarounds are needed on unsolved vulnerabilities but not on ones for which patches are available.

Additionally, Intel highlights our implementation strategy as still being secure enough to even disable the patch in their [technical documentation of the issue](https://software.intel.com/content/www/us/en/develop/articles/software-security-guidance/technical-documentation/special-register-buffer-data-sampling.html), under "mitigations":

> If usage of RDRAND/RDSEED is very frequent, system administrators may choose to opt out of the mitigation if any of the following are true:
> [..]
> 3. RDRAND or RDSEED output is securely mixed with other sources of entropy before use (for example, in the Linux* getrandom(2) system call). 

Since that describes what we do - I'd need some hard data to make a decision to not pull this in. Or a pointer of where to get said data. Let me know.

ReverseControl: @patricklodder On your first point, I agree. I don't think flushing of any kind is really needed, per your reasoning. On your second point, yes, I saw that. Furthermore, there is even an MSR that allows you to disable the microcode patch. I forget the intel doc, but it can be found if need be.

I was addressing @edtubbs as I myself ask for documentation when I do not have firm grounds to make a technical decision. My response was just for present and future reference.

As far as I am concerned, this is ready to go.

patricklodder (commented):
> As far as I am concerned, this is ready to go.

We need a rebase/squash on this into a nice singular commit. Please squash mine in it to, if you want, you can use a `Co-authored-by:` line in your commit message to describe which code of mine you pulled.

I cannot give the final ACK on this now because I contributed code but I'm testing this as we speak.

src/random.cpp
edtubbs: @patricklodder, @ReverseControl mentioned in https://github.com/dogecoin/dogecoin/pull/2464 this was a back port from 1.21; it's also been in bitcoin core since 2019, prior to this CVE https://github.com/bitcoin/bitcoin/blob/7fcf53f7b4524572d1d0c9a5fdc388e87eb02416/doc/release-notes/release-notes-0.18.0.md
I'd be inclined to recommend 2; however if the consensus is the value outweighs the risk then I'm inclined to agree.

ReverseControl: @edtubbs What I implement here is not the same as that backport you reference which I started. This one uses the hardware RNG directly; in addition to the OS dependent implementation of random, which may or may not use RDSEED. Also, I added 64 bytes for each source, instead of the previous  32, which that backport still uses.

That backport deals better with different operating systems, as they have some quirks for getting random bits.

edtubbs: NIST SP 800-90B recommends "twice" the entropy of the underlying block of a cryptographic hash function be used as input in order to have an output hash having fully saturated entropy capacity; in this case with SHA-512 the combination of two 64-byte sources is sufficient seed entropy.

ReverseControl: @edtubbs That being the case I think what we have is a perfect balance. The first two sources would saturate entropy at 128 bytes, and RDSEED where available would add an additional 64 byte block. 

Just for reference, could you please provide the exact reference to that fact.

edtubbs: Actually, it clearly exceeds the spec.  Given that, consider it should be considered a build option as opposed to enabled by default? 
For the reference I mentioned above n 90C, line 349, https://csrc.nist.gov/csrc/media/publications/sp/800-90c/draft/documents/sp800_90c_second_draft.pdf

ReverseControl: "including a requirement that the entropy-source output that is provided as input to the conditioning 348 
function  have AT LEAST twice  the  amount  of  entropy  as  the  number  of  bits  that  are  349 
produced  as  output  from  the  conditioning  function"

Says at least. Clearly within the spec. This added security comes at little performance cost. I see no reason to make it optional.

edtubbs: openssl would claim that's not necessary, it doesn't seem necessary to make it required

ReverseControl: The claim is that I support its inclusion, not that it is necessary. I defer to the core devs as well.

edtubbs:
https://software.intel.com/content/www/us/en/develop/blogs/changes-to-rdrand-integration-in-openssl.html
According to Intel, OpenSSL is using RDSEED as an entropy source to seed the software PRNG in v1.0.2..
https://github.com/openssl/openssl/blob/1c0eede9827b0962f1d752fa4ab5d436fa039da4/providers/implementations/rands/seeding/rand_cpu_x86.c#L29

ReverseControl:
@edtubbs Thanks for that reference. That Intel document supports the option to implement RDSEED ourselves, given OpenSSL does not gather entropy from that instruction by default.

edtubbs:
Actually, it says OpenSSL uses the RDSEED instruction by default if available with RDRAND starting in v1.0.2, and the comments in that version are consistent.

ReverseControl:
@edtubbs My final comment on this matter: I vote for the inclusion of this code. We maintain this code and can guarantee that sources from the hardware, if available in hardware and sufficient in bytes at running time, will always be used. We will not depend on other projects implementation/usage not even the OS kernel.  Virtually no performance cost. Per NIST standard, compliant. Small enough to review carefully. 

Per OpenSSL project, note this was written 3 years after the intel doc, (https://www.openssl.org/blog/blog/2017/08/12/random/):

```
They’re relatively well-commented, but the implementation could to change in the future. 
(referencing entropy sources.)
```

@michilumin @rnicoll will have the final word on this.

patricklodder:
@ReverseControl could you please do the rebase/squash on this branch?

If you haven't the remote to this repo already:
```
git remote add upstream https://github.com/dogecoin/dogecoin.git
```

then:
```
git fetch upstream
git rebase -i upstream/1.14.5-dev #and squash all the commits into one
git push -f origin 1.21.5-dev-improve_entropy_for_key_generation
```

then this PR should show a clean commit tree and becomes mergeable.

patricklodder:
Pushing to 1.14.6 as this is stalled on rebase.

ReverseControl:
@patricklodder Done. Do check the rebase.

patricklodder (changes requested):
First off, my apologies for having this fall off my radar. I simply missed this one when I was building up the review board. 

Second, I have done some research and unfortunately we cannot release this as-is, because of a problem with some AMD processors that have the needed capabilities disabled, see https://github.com/weidai11/cryptopp/issues/924

As a result, for this to make it into a released binary, we need runtime detection and not just compile-time detection, because anything less can lead to failures in the field (a lot of shibes use older hardware to run a node that they otherwise wouldn't use anymore.)

I recommend we take the following path to that:

1. Change the disable flag to an enable flag in `configure.ac`, and make the default `no`. That makes this an experimental feature, so we add the config to the experimental CI.
2. When we get to the point that we have a generic framework for the runtime cpu capability detection as mentioned (and a little bit discussed) under #2773, we can integrate that for this functionality too and decide to move this out of experimental state, into production state.

I would not recommend to create a custom runtime detection right now, simply because we probably won't have time for thorough testing before a 1.14.6 release, and I think it would be a shame if we'd skip this PR once more.

CryptoUnchained (approved):

PR #2487, opened 2021-08-22
Fix compiler warnings

chromatic:
This PR fixes a handful of compiler warnings found when testing #2484.

While editing, I found and cleaned up a couple of whitespace inconsistencies in a separate commit at the end.


chalermchai29 (changes requested):
src/dogecoin-fees.cpp

patricklodder (approved):
ACK. Those are very useful bug fixes. Thank you!

PR #2522, opened 2021-09-03
Merge Bitcoin Core 0.21.2-rc2

rnicoll:


Ufakeimdope:
I want too know my total balance how is that possible on any chain because
ik I have been using my hash and my device or my something because it's
saying bit coins are on my device and it's annoying I need some help

ReverseControl:
@Ufakeimdope Submit a new issue, give detail of your situation and ask your question. This is not the place for that.

chromatic:
I skimmed the commits to see what looks like it came from Bitcoin upstream and read the unique commits. Everything looks sensible in terms of provenance.

I checked out this branch, then built it and ran tests. Everything looks fine except for:

```bash
95/190 - rpc_signrawtransaction.py --descriptors failed, Duration: 149 s
```

The relevant error lines may be:

```bash
test_framework.authproxy.JSONRPCException: 'generatetoaddress' RPC took longer than 30.000000 seconds. Consider using larger timeout for calls that take longer to return. (-344)
2021-09-19T21:25:06.622000Z TestFramework (INFO): Stopping nodes
2021-09-19T21:25:06.622000Z TestFramework.node0 (ERROR): Unable to stop node.
```

I ran that test again on its own and it passed within 10 seconds, so that may reflect additional load on my machine.

rnicoll:
> 
> 
> I skimmed the commits to see what looks like it came from Bitcoin upstream and read the unique commits. Everything looks sensible in terms of provenance.
> 
> I checked out this branch, then built it and ran tests. Everything looks fine except for:
> 
> ```shell
> 95/190 - rpc_signrawtransaction.py --descriptors failed, Duration: 149 s
> ```
> 
> The relevant error lines may be:
> 
> ```shell
> test_framework.authproxy.JSONRPCException: 'generatetoaddress' RPC took longer than 30.000000 seconds. Consider using larger timeout for calls that take longer to return. (-344)
> 2021-09-19T21:25:06.622000Z TestFramework (INFO): Stopping nodes
> 2021-09-19T21:25:06.622000Z TestFramework.node0 (ERROR): Unable to stop node.
> ```
> 
> I ran that test again on its own and it passed within 10 seconds, so that may reflect additional load on my machine.

Yeah, I had a similar experience with that test failing in batch but passing individually. I might throw it through the loop a few more times just to make sure it's consistent, but generally I use CI as the final one it has to pass.

patricklodder:
@rnicoll Am I correct to assume that there were no conflicts with the merge, and that all changes on top of bitcoin's commits are in the 2 commits of yours at the top of the chain (displayed here at the bottom)?

rnicoll:
@patricklodder Had a few conflicts. I resolved conflicts in the merge commit, and test fixes in the follow-on commits.

Conflicts are:

```
.cirrus.yml
depends/packages/boost.mk
src/chainparams.cpp
src/test/data/key_io_invalid.json
src/test/data/key_io_valid.json
src/test/versionbits_tests.cpp
test/functional/test_framework/segwit_addr.py
test/functional/wallet_labels.py
```

* `.cirrus.yml` was because we disabled a broken test, which was since fixed.
* `depends/packages/boost.mk` was the download path. I've used the Bitcoin Core update.
* `src/chainparams.cpp` was because they enabled Taproot. I've reverted those changes.
* `src/test/data/` was because they rebuilt those data files. I rebuilt then and included that in the merge commit, because neither old nor new data was correct.
* `src/test/versionbits_tests.cpp` our block time differs from Bitcoin, I've used our blocktime but otherwise retained their changes.
* `test/functional/test_framework/segwit_addr.py` they added a further test value, and Git was worried our changes interacted. I've resolved by adding their test value (around line 133).
* `test/functional/wallet_labels.py` their changes included addresses, which I've rewritten to Dogecoin equivalents.

Hope that helps

patricklodder:
> Hope that helps

Thanks! It does!

Looking at there being some conflicts, and pre-merging other PRs could make even more conflicts, shall we do the following order?

1. This PR (#2522)
2. #2560 and #2561 (I am considering just reviewing the latter, the former doesn't make much sense without it (thoughts ?!?))
3. #2562 (I tested 2561 + 2562 a bit and I think that this for now will have to do)

And keep #2524 for later per our discussion in there.

When would you want to start with #2533 to #2545?

rnicoll:
With #2533 to #2545 I've encouraged that they attempt to merge the improvements upstream, so we pre-empt further problems.

Definitely my preferred order is we merge this first, then anything else in the 1.21 range.

patricklodder (approved):
Two non-blocking issues:

1. I'm still getting this error when building with depends using a focal container:

```
FAIL: qt/test/test_dogecoin-qt
==============================

./qt/test/test_dogecoin-qt: error while loading shared libraries: libfontconfig.so.1: cannot open shared object file: No such file or directory
FAIL qt/test/test_dogecoin-qt (exit status: 127)
```

I'll try to find out why, and raise an issue or PR - depending on what I find.

2. DNS seeding on mainnet is currently not working - but I think that's an issue with the seeds not supporting the service flag discriminators rather than with the code.

Other than that, this syncs headers until 370000, which is as expected.

ACK.

PR #2524, opened 2021-09-04
Clean up wallet_groups.py

rnicoll:
Minimise changes to wallet_groups.py to make it Dogecoin compatible. Previously fee values were tweaked and addresses set to bech32 everywhere, now it uses bech32 (rather than Dogecoin's default of "legacy") addresses by default and this removes the need for fee value changes.

### DIFF
--- a/test/functional/wallet_groups.py
+++ b/test/functional/wallet_groups.py
@@ -17,11 +17,11 @@ class WalletGroupTest(BitcoinTestFramework):
         self.setup_clean_chain = True
         self.num_nodes = 5
         self.extra_args = [
-            [],
-            [],
-            ["-avoidpartialspends"],
-            ["-maxapsfee=0.00002719"],
-            ["-maxapsfee=0.00002720"],
+            ["-addresstype=bech32"],
+            ["-addresstype=bech32"],
+            ["-avoidpartialspends", "-addresstype=bech32"],
+            ["-maxapsfee=0.00002719", "-addresstype=bech32"],
+            ["-maxapsfee=0.00002720", "-addresstype=bech32"],
         ]
         self.rpc_timeout = 480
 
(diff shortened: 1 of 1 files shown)


rnicoll:
Bitcoin equivalent for reference: https://github.com/bitcoin/bitcoin/blob/v0.21.1/test/functional/wallet_groups.py

patricklodder:
> to make it Dogecoin compatible

How are bech32 addresses Dogecoin compatible? Per https://github.com/dogecoin/dogecoin/commit/261d38e0feda61afd6882f166db213a0ac8672b5 we use "legacy" by default, not bech32, because we would need to deploy bitcoin's segwit for that and how does that relate to [the challenges received from contributors regarding the consensus plan](https://github.com/dogecoin/dogecoin/discussions/2264)?

rnicoll:
It's enabled on regtest, but not main or test. The alternative is a much more invasive changeset, and would risk that we introduce a break to the segwit code without realizing it. That's an issue if we want to look at enabling it post 1.21.0.

You can check the disabling in chainparams.cpp

polska333:
This ain't fair I'm losing sleep over this and everything all my doge and
xrp are gone .. I'm disabled I eat through a damn tube I died once already
tell me they don't know when it's going to happen again my cryptos all I
had to help me through my illnesses. Then they have it all just White away
from you this is like downloaded that new core . It has to be someone of
someone I can contact people who knew how hard I worked to get up there

rnicoll:
Realized this perhaps isn't obvious - the original used BECH32 addresses, this is resetting it back to equivalent to the original known-good test.

rnicoll:
Been experimenting with this more over on the fees branch - definitely can replace all the values in the rest with legacy address versions, if that's preferable. It needs redoing once we update to Dogecoin fees, anyway.

polska333:
I am way too upset because all my doge is missing it happened when I did
the upgrade I don't know what you guys did with all my dogecoin but it is
not right I'm just able I need through a feeding tube I work very hard to
earn my coin this is not right at all something needs to be done

patricklodder:
> Been experimenting with this more over on the fees branch - definitely can replace all the values in the rest with legacy address versions, if that's preferable. It needs redoing once we update to Dogecoin fees, anyway.

That would be my preference, because a bech32-based test has 0 value if there is no bech32 on mainnet.

patricklodder:
@polska333 if you have an issue, please open an issue.

rnicoll:
Closing and will adapt figures to Dogecoin fees later instead.

PR #2525, opened 2021-09-04
Replace Bitcoin with Dogecoin in translations

chromatic:
This replaces the string "Bitcoin" with "Dogecoin" in sources, case-insensitively, and replaces the translated strings where I could do so appropriately. I don't speak or write any of these languages natively, so the translations could use extra review.

Fixing the source strings at least lets us translate these messages at all.


choroba:
Lines 171 + 172 in the Slovak file still contain "BITCOIN".

chromatic:
> Lines 171 + 172 in the Slovak file still contain "BITCOIN".

I filed #2526 for that.

chromatic:
> Lines 171 + 172 in the Slovak file still contain "BITCOIN".

Oh, I misunderstood you. Fixed now. Thanks!

PR #2533, opened 2021-09-05
Deepsource Fix: Logging is not lazy 

ReverseControl:
The logging statement has the call of the form logging.(format_string % (format_args...)). For such calls, it is recommended to leave string interpolation to the logging method itself and be written as logging.(format_string, format_args...) so that the program may avoid incurring the cost of the interpolation in those cases in which no message will be logged. For more details, see PEP 282.


rnicoll:
I think it's a hard case to make that this is worth it for us, performance of the tests isn't a big deal, and this will make it harder to merge upstream changse. Easiest solution would be can you try raising against Bitcoin Core?

PR #2534, opened 2021-09-05
Deepsource Fix: Bound instance not used. Decorate as staticmethod.

ReverseControl:
The method doesn't use its bound instance. Decorate this method with @staticmethod decorator, so that Python does not have to instantiate a bound method for every instance of this class thereby saving memory and computation


CryptoUnchained (approved):

CryptoUnchained (approved):

CryptoUnchained (commented):
looks fine here

CryptoUnchained (approved):

CryptoUnchained (approved):

PR #2537, opened 2021-09-05
Deepsource Fix: Unnecessary use of iterable comprehension.

ReverseControl:
It is unnecessary to use a comprehension just to loop over the iterable and create a list/set/dict out of it. Python has a specialized set of tools for this task: the list/set/dict constructors, which are faster and more readable.


ZachOlivier (approved):
References:
https://stackoverflow.com/questions/29356846/python-list-vs-list-comprehension-building-speed

Python Docs:
https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions

PR #2538, opened 2021-09-05
Deepsource Fix: Comprehension not required. Allow evaluation short-circuiting.

ReverseControl:
The built-in function being used does not require comprehension and can work directly with a generator expression.

Using a generator expession within these functions is faster than using a comprehension.

    all
    any
    enumerate
    iter
    itertools.cycle
    itertools.accumulate

Note: The inbuilt functions all() and any() in python also support short-circuiting (evaluation stops as soon as the overall return value of the function is known), but this behavior is lost if you use comprehension. This affects performance.


ReverseControl:
@patricklodder Can we trigger the failing test on this as well? Curl error:

```curl: (92) HTTP/2 stream 1 was not closed cleanly: REFUSED_STREAM (err 7)```

rnicoll:
I'm not sure how much we should worry about performance of the tests, but looks like it should be good cleanup!

rnicoll (approved):

PR #2539, opened 2021-09-05
Deepsource Fix: Identity check for singleton objects is better than equality check.

ReverseControl:
Comparisons to the singleton objects, like True, False, and None, should be done with identity, not equality. Use is or is not.

Identity checks are faster than equality checks. Also, the equality checks can result in unintended behaviour in some cases.


ZachOlivier (approved):
References:
https://codereview.doctor/features/python/best-practice/avoid-equality-checking-boolean-none
and
https://stackoverflow.com/questions/25317175/why-comparisons-to-singletons-like-none-should-always-be-done-with-is-or-is-not

PR #2541, opened 2021-09-05
Deepsource Fix: Drop redundant use of elif/else statements.

ReverseControl:
The use of else or elif becomes redundant and can be dropped if the last statement under the leading if / elif block is a return statement. In the case of an elif after return, it can be written as a separate if block. For else blocks after return, the statements can be shifted out of else. Please refer to the examples below for reference.

Refactoring the code this way can improve code-readability and make it easier to maintain.


ReverseControl:
@patricklodder Can we trigger only the failing test please? It got a curl error:

``` curl: (22) The requested URL returned error: 503  ```

PR #2542, opened 2021-09-05
Deepsource Fix: Implicit enumerate calls found.

ReverseControl:
Using range(len(...)) is not pythonic. Python does not have not index-based loops. Instead, it uses collection iterators.

Python has a built-in method enumerate which adds a counter to an iterable. Using this, you can access the counter and the value from the iterable at the same time. It is therefore recommended to replace range(len(...)) with enumerate(...).


rnicoll:
I'm cautious of this, because it will make it harder to merge upstream changes. Potentially consider raising against bitcoin/bitcoin instead?

ReverseControl:
@rnicoll At some point we should get out of the shadow of bitcoin. Even in the code.

PR #2546, opened 2021-09-05
Refresh English translation file

rnicoll:


rnicoll:
@langerhans I've created this with Qt's `lupdate` tool, can you let me know if the compiled result in German is sane, please?

edtubbs (approved):

patricklodder (approved):
utACK with the note that as long as we don't close GUI development for 1.14.5, this isn't final. But it doesn't matter, we'll just do it twice.

PR #2561, opened 2021-09-10
1.21 auxpow header

rnicoll:
Add data type to hold a merge-mined header from another blockchain, along with details of how to prove the merge-mined header is intended for the Dogecoin chain.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -107,6 +107,7 @@ BITCOIN_CORE_H = \
   addrdb.h \
   addrman.h \
   attributes.h \
+  auxpow.h \
   banman.h \
   base58.h \
   bech32.h \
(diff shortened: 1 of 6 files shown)


patricklodder (commented):
`src/test/auxpow_tests.cpp` is added here but is never compiled as it isn't added to the test suite. If I add it to `src/Makefile.test.include`, it errors out:

```
  CXX      test/test_dogecoin-auxpow_tests.o
test/auxpow_tests.cpp:14:10: fatal error: rpc/auxpow_miner.h: No such file or directory
   14 | #include <rpc/auxpow_miner.h>
      |          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

rnicoll:
AuxPoW tests shouldn't have been in this, it doesn't have enough functionality for them to run. Moved to the next PR.

patricklodder (commented):
I think the `src/test` should go into your extracted commit with the tests. Everything else is fine. 

I did some more extensive testing on this (mostly: testing the upgrade) but this will currently purge all blocks past auxpow height (but no longer be much confuse about parsing the block files)

rnicoll:
@patricklodder Removed the rest of the `src/test` changes and pushed

patricklodder (approved):
tested & code review ACK.

Unit & functional tests come with a later PR

PS: thanks for getting rid of most of that awkward Namecore formatting, it was hurting my eyes ❤️

Issue #1604, opened 2019-08-05
settxfee no longer working?


Joter271:
Dogecoin Core v1.14.0

Settxfee RPC call doesn't set fee. There is no error returned by RPC call, but it doesn't change transaction fee. Same RPC call works fine with  Dogecoin Core v1.11.

EDIT:
After some more digging I have found that paytxfee is actually set correctly, but transaction is always sent with fee 1 DOGE, no matter the fee size we set. We use sendtoaddress RPC call.

Here is  getinfo output:￼
getinfo￼
{
  "version": 1140000,
  "protocolversion": 70015,
  "walletversion": 130000,
  "balance":  xxxxxx,
  "blocks": 2838653,
  "timeoffset": -1,
  "connections": 8,
  "proxy": "",
  "difficulty": 6321115.617799241,
  "testnet": false,
  "keypoololdest": 1564688963,
  "keypoolsize": 100,
  "paytxfee": 5.40000000,
  "relayfee": 1.00000000,
  "errors": ""
}

EDIT2:
Temporal solution was setting mintxfee paramater in config. This is only way of increasing sendtoaddress transaction fee...

Joter271:
So, any update on this?

ReverseControl:
@Joter271 This issue might be fixed in 1.14.4. Could you please test it and post the update.

ReverseControl:
@Joter271 this issue was addressed in 1.14.4. Transaction relay fee along with several other values were fixed to 1 DOGE / Kb, where transaction size was rounded up. All one-to-one transactions take less than 1KB, so previously they all got rounded up to 1KB, so their cost became 1 DOGE. And because of this, and some technical issues (bugs) which we have now resolved (See https://github.com/dogecoin/dogecoin/pull/2446), the system actually counts the number of bytes per transaction without rounding (See https://github.com/dogecoin/dogecoin/pull/2485/),  and applies the fees that you set correctly. The next release will decrease the default fee values (See https://github.com/dogecoin/dogecoin/discussions/2347).

For reference:

Dogecoin Core 1.14.5 Default Fees


Parameter | Current value | Proposed value
|-- |--|--|
|MIN_RELAY_FEE | 100,000,000 | 100,000
|DEFAULT_FALLBACK_FEE | 100,000,000 | 1,000,000
|DEFAULT_TRANSACTION_FEE | 0 | 1,000,000
|DUST_RELAY_TX_FEE | 100,000,000 | 100,000
[...]

PR #2566, opened 2021-09-12
Hide smart fee UI elements

rnicoll:
Hide smart fee UI elements, as they do not make sense for Dogecoin. Dogecoin uses the defined minimum in virtually all transactions, rather than smart fees, and this Bitcoin-focused UI does not match the expected workflow.

### DIFF
--- a/src/qt/forms/sendcoinsdialog.ui
+++ b/src/qt/forms/sendcoinsdialog.ui
@@ -759,6 +759,7 @@
             </item>
            </layout>
           </item>
+          <!--
           <item>
            <widget class="QLabel" name="fallbackFeeWarningLabel">
             <property name="toolTip">
@@ -778,6 +779,7 @@
             </property>
            </widget>
           </item>
+          -->
           <item>
            <spacer name="horizontalSpacer_4">
             <property name="orientation">
@@ -896,14 +898,20 @@
                 <layout class="QHBoxLayout" name="horizontalLayoutFee8">
                  <item>
                   <widget class="QCheckBox" name="checkBoxMinimumFee">
+                   <!--
                    <property name="toolTip">
                     <string>Paying only the minimum fee is just fine as long as there is less transaction volume than space in the blocks. But be aware that this can end up in a never confirming transaction once there is more demand for dogecoin transactions than the network can process.</string>
                    </property>
+                     -->
                    <property name="text">
                     <string/>
                    </property>
+                   <property name="checked">
+                    <bool>true</bool>
+                   </property>
                   </widget>
                  </item>
+                 <!--
                  <item>
                   <widget class="QLabel" name="labelMinFeeWarning">
                    <property name="enabled">
@@ -920,6 +928,7 @@
                    </property>
                   </widget>
                  </item>
+                   -->
                  <item>
                   <spacer name="horizontalSpacer_2">
                    <property name="orientation">
@@ -944,8 +953,8 @@
                  <property name="text">
                   <string>Recommended:</string>
                  </property>
-                 <property name="checked">
-                  <bool>true</bool>
+                 <property name="enabled">
+                  <bool>false</bool>
                  </property>
                  <attribute name="buttonGroup">
                   <string notr="true">groupFee</string>
@@ -973,6 +982,9 @@
                 <widget class="QRadioButton" name="radioCustomFee">
                  <property name="text">
                   <string>Custom:</string>
+                 </property>
+                 <property name="enabled">
+                  <bool>false</bool>
                  </property>
                  <attribute name="buttonGroup">
                   <string notr="true">groupFee</string>
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -594,7 +594,8 @@ void SendCoinsDialog::updateFeeSectionControls()
     //mlumin: 5/2021 - this label actually gates the 'slider and smart fee' functionality, so turn it off for dogecoin.
     ui->confirmationTargetLabel ->setEnabled(ui->radioSmartFee->isChecked());
     ui->checkBoxMinimumFee      ->setEnabled(ui->radioCustomFee->isChecked());
-    ui->labelMinFeeWarning      ->setEnabled(ui->radioCustomFee->isChecked());
+    // Dogecoin: We want minimum fees, don't warn about it
+    // ui->labelMinFeeWarning      ->setEnabled(ui->radioCustomFee->isChecked());
     ui->radioCustomPerKilobyte  ->setEnabled(ui->radioCustomFee->isChecked() && !ui->checkBoxMinimumFee->isChecked());
     ui->radioCustomAtLeast      ->setEnabled(ui->radioCustomFee->isChecked() && !ui->checkBoxMinimumFee->isChecked() && CoinControlDialog::coinControl->HasSelected());
     ui->customFee               ->setEnabled(ui->radioCustomFee->isChecked() && !ui->checkBoxMinimumFee->isChecked());
@@ -659,11 +660,12 @@ void SendCoinsDialog::updateSmartFeeLabel()
                                                                 std::max(CWallet::fallbackFee.GetFeePerK(), CWallet::GetRequiredFee(1000))) + "/kB");
         ui->labelSmartFee2->show(); // (Smart fee not initialized yet. This usually takes a few blocks...)
         ui->labelFeeEstimation->setText("");
-        ui->fallbackFeeWarningLabel->setVisible(true);
-        int lightness = ui->fallbackFeeWarningLabel->palette().color(QPalette::WindowText).lightness();
-        QColor warning_colour(255 - (lightness / 5), 176 - (lightness / 3), 48 - (lightness / 14));
-        ui->fallbackFeeWarningLabel->setStyleSheet("QLabel { color: " + warning_colour.name() + "; }");
-        ui->fallbackFeeWarningLabel->setIndent(QFontMetrics(ui->fallbackFeeWarningLabel->font()).width("x"));
+        // Dogecoin: We don't care we're using fallback fees
+        // ui->fallbackFeeWarningLabel->setVisible(true);
+        // int lightness = ui->fallbackFeeWarningLabel->palette().color(QPalette::WindowText).lightness();
+        // QColor warning_colour(255 - (lightness / 5), 176 - (lightness / 3), 48 - (lightness / 14));
+        // ui->fallbackFeeWarningLabel->setStyleSheet("QLabel { color: " + warning_colour.name() + "; }");
+        // ui->fallbackFeeWarningLabel->setIndent(QFontMetrics(ui->fallbackFeeWarningLabel->font()).width("x"));
     }
     else
     {
@@ -671,7 +673,7 @@ void SendCoinsDialog::updateSmartFeeLabel()
                                                                 std::max(feeRate.GetFeePerK(), CWallet::GetRequiredFee(1000))) + "/kB");
         ui->labelSmartFee2->hide();
         ui->labelFeeEstimation->setText(tr("Estimated to begin confirmation within %n block(s).", "", estimateFoundAtBlocks));
-        ui->fallbackFeeWarningLabel->setVisible(false);
+        // ui->fallbackFeeWarningLabel->setVisible(false);
     }
 
     updateFeeMinimizedLabel();


rnicoll:
Functionally this is an alternative to #2521, with the intent 2521 could be applied in a later release after more design work.

rnicoll:
pay custom fee
pay only minimum

edtubbs (approved):
Looks good; I've rebuilt and tested these changes as well.

patricklodder (changes requested):
This proposal is inconsistent with changes proposed in #2564, and basically leaves the choice between 2 evils because in the current state of #2521 you propose something that inflates fees to incredible levels, yet here you propose people type in amounts (more error prone than a pre-set slider) This is not awesome, at all.

What is the goal?

- no "smart fees"? That's something @michilumin and you were working on the past 4-5 months. What's the status of that? You've been telling the whole world through social media that "the code is ready", so where is it? Hiding UI elements to buy time for research seems a bit contrary to these prior claims.
- Do we encourage smart fees per your proposal in #2564? Then how does this (or #2521) integrate with that? I don't see that integration anywhere.

If you want to buy time, how much time? Who is working on the UI design you mention and how long do they need to finish that?

rnicoll:
This defaults to using minimum fees, it just retains the ability to modify them.

* #2521 was a UI prototype, which mostly exposed that the thoughts on how the UI with a slider should look weren't fully baked. I'll close it in a minute so it doesn't clutter things up.
* #2564 solves the issue that some wallet providers want to use the smart fees, by having the values produced at least be vaguely sane. It also means we don't have code that doesn't do anything, as currently we're not capturing any usable data.

The references to code being ready were about 114.4 (certainly that's my understanding of the question posed). Fundamentally we have 1.14.5 doing working fees, just the user experience isn't great (hence this patch). On #2564, I think it's a nice bug fix, but if fixing the code behind the scenes causes problems we can force estimatesmartfees to produce fixed values instead.

Michi's been working on UI designs. Functionally this is my proposed solution, it removes an option that does nothing (it works correctly, because our recommended fees are always 0.01 DOGE, but it's not _helpful_), and potentially then look at whether 1.14.6 needs the slider later.

patricklodder:
> #2521 was a UI prototype, which mostly exposed that the thoughts on how the UI with a slider should look weren't fully baked. I'll close it in a minute so it doesn't clutter things up.

Alright. So on 1.14.4, we [reversed](https://github.com/dogecoin/dogecoin/pull/2446) the previous code from d0133b2 where (inadvertently - both you and @langerhans have explained to me on multiple occasions that these choices were made under release pressure) user-defined fees were overridden with developer enforced fees. This means that between 1.14.0 and 1.14.4, meaning for about 2.5 years, the slider did not work at all, as for example #1604 and #1691 highlighted. During all these years, while a fixed fee was enforced on the network and the slider didn't do anything, it did not get removed, but now that fees are variable again, we're removing the slider in favor of only having an input field, instead of retaining an arguably user-friendly way to deal with variable fees. This change would have made sense back when we released 1.14.0, 1.14.2 or 1.14.3, but now, it is at the very least brow-raising, especially when the reason is implied to be "release pressure", once more.

Additionally, I've [recommended a set of slider values that cover 4 different scenarios](https://github.com/dogecoin/dogecoin/pull/2521#issuecomment-913811448), but you have not responded to that at all. Is that because you disagree with my suggestion, disagree with the scenarios, or have no time to evaluate? You say more research needs to be done, but you're not having the discussion, instead you have now closed that discussion off.

Subsequently, you suggested to remove the slider instead, asking for opinions. I then gave you an opinion against disabling the slider but instead just doing the work that's needed, which you did not address either, again, just closing the discussion off before it even started. Why?

> some wallet providers want to use the smart fees, by having the values produced at least be vaguely sane. It also means we don't have code that doesn't do anything, as currently we're not capturing any usable data.

Fair enough. I'll take the discussion on that one there.

> (it works correctly, because our recommended fees are always 0.01 DOGE)

Is it safe then to assume that your position remains that smart fees should not be used and recommended fees are static? How many relay nodes and how many miners have upgraded to allow lower fees according to your statistics that this is possible? And what is the source of those statistics?

For the relay side, here are my Shibetoshi peer statistics from 17 of my publicly accessible mainnet nodes that as of writing connect to 2670 unique peers, 2642 of which are Dogecoin Core (or imitate it), [collected using dogecoin-cli](https://gist.githubusercontent.com/patricklodder/7aba4113aab9da68e65cfd1a7f3bb0b3/raw/cda0f8bf98926aa73edc684017549756a8894f94/unique_shibetoshi.sh):

```
    num version
   ---- --------
      1 Shibetoshi:1.21.0
     22 Shibetoshi:1.10.0
     49 Shibetoshi:1.14.0
    619 Shibetoshi:1.14.4
    637 Shibetoshi:1.14.2
   1314 Shibetoshi:1.14.3
```

***This implies that we're still under 30% deployment, even when counting 1.10.0 and 1.14.0 as friendly to lower fees.***

Over the last 4000 blocks, I have also observed, [using the fee tools that I created for you all last year](https://github.com/patricklodder/fee-stats) that f2pool is the only miner currently processing lower fee transactions. Thank you f2pool for being awesome! No other pool has mined transactions < 1 DOGE/kb rounded in that period, implying that either all other miners have not upgraded, or they are now requiring fees equal to or larger than the previously enforced ones. As that pool sees 20-23 percent of blocks on good days, 15-17% of blocks on bad days, this is far from ideal. Finalizing an advisory and pushing a hardcoded default fee out based on a single mining pool adapting to the new network rules is in my opinion risky and premature, because there is then a single point of failure in the assertion that lower fees get mined.

So once again, I think there is time. And please, if anyone disagrees, could you please be so kind to share your objections here?

> Michi's been working on UI designs.

Is there anything to share or has this just started right now? I understand that it's hard times when every developer in your organization is burned out according to your press release, so maybe sharing the designs, even when unfinished, will trigger shibes to help pulling this over the finish line? I'd personally help for sure, and I'd rather spend time doing that, than having to argue protection of this coin with you all the time because you're taking dangerous short cuts. Break the cycle? That way, everyone wins, especially shibes.

rnicoll:
So the mismatch on #2521 is that it wasn't my idea, I just mocked up code to demonstrate what it would look like. I had thought your suggestion was how we might save the PR, not it was mostly the right path but just needed refining. If the answer is you are happy with #2521 with the fee values changed (and noting I believe the labels aren't final either), I'm happy to reopen and revise it, I just hadn't expected it would get through review any time soon (hence it never made it out of draft status).

On smart fees - I definitely want to keep them away from the user, in particular having a difference in miner inclusion and minimum relay values means they can't be guaranteed to work. **I am uncertain** on how to resolve the issue with wallet providers. Main options I see are using static values, removing the RPC command to force use of `estimatefee`, or modifying the smart fee estimates to specific Doge-friendly values (i.e. ensure the amount is never below the block inclusion minimum). I would appreciate thoughts there, because the main consideration is what others would accept.

I'm using https://blockchair.com/dogecoin/nodes for node data, although that will be limited to nodes which are reachable, it won't include nodes which can only connect out. It shows 770 1.14.4 nodes, 482 1.14.3 nodes and 234 1.14.2 nodes as of the time of writing. We know F2Pool have upgraded, because we've talked to them about it, and at least one other miner (because low fee transactions were being mined before F2Pool upgraded). I've also been doing live network tests with low fee transactions and while they take a couple of blocks to confirm, we know they do confirm.

On UI designs - #2521 was part of that, although as said I think the labels were not yet final, and there was talk of adding in an easter egg. I know there's a desire to have a few more aesthetic tweaks on it, I don't believe they're imminent.

Path forward:

* If you're mostly happy with #2521, I'm happy to reopen, apply your changes, and send that to wider review. That would then replace this PR.
  * There's a third UI option I've been pondering, of simply radio buttons for old/new fees. I don't think that's what anyone's expecting, but raising it in case you do want to go "Yes, that one."
* I'd appreciate thoughts on what `estimatesmartfees` should produce in 1.14.5. If we are using pre-defined values on the slider, it might make sense that #2564 is modified so the returned smart fee is always one of those values, which would give the user and wallet providers similar results.

patricklodder:
> So the mismatch on #2521 is that it wasn't my idea, I just mocked up code to demonstrate what it would look like.

I think that in such cases it's better to actually have the discussion rather than just leaving and closing it. To me, fixing the slider to something meaningful and usable is good - better than just having a text field - and as I said if there's some humor in there it's cool, but it shouldn't be misleading or dysfunctional. If things need more thought then we just need to work through it, and it would be great if you could write that in your PR comment next time! A simple note saying "This is just a mock up to get the discussion started" means that instead of focusing on reviewing code, we discuss the concept.

As for the comments on that PR - and others: because unfortunately your org has been showing the behavior - most importantly on #2485, but on others too - that in the past we all agreed we should be wary about: to have one org member approve and merge the PRs of the other, you create the situation where, wherever anyone formally or informally associated with your organization raises a PR, there is no time to approach things calmly and well thought out, but instead, force people that actually understand the code and think about implications of a change to challenge in public ASAP or else risk that you make merge arrangements internally. I'm not only referring to published members of the foundation either, as you have been forming private, segregated groups of developers that you control who are quick to do you favors, put their name under your code even when they don't understand it and/or the quality is terrible, and "approve" your PRs without thought or vision. The problem is that if you don't think your PRs through, and then reviewers don't take the time to think things through because they are focused doing you, your org, or its billionaire sponsors favors, and then on top you create release pressure because you want to get a press release out, then you, your org and your following - as well intended as it may be - are becoming a threat and you leave me no choice but to engage with you accordingly.

It doesn't have to be that way, your foundation can always "JDart" their way out of their governance claim and let the community govern, as was promised early April 2014 by Jackson Palmer. I'm sure your legal team knows how to take that path - they invented the concept. We can then define a workable process that protects the coin and shibes, and move forward in a more harmonious way. I am confident everyone has good intentions, but you also have to **_do_** good... it's your motto after all.

> I just hadn't expected it would get through review any time soon

Of course it would get through review if the values and labels make sense and aren't misleading or dysfunctional. I think the code is fine at a glance, just the parametrization is bad and the labels are misleading. I think it's important that devs think about usage scenarios and problems that the end user can run into, and solve these to the best of their abilities, because if the values are hardcoded (which is the proposal) then shibes running that software will be stuck with these for a longer time, so thinking ahead a bit is not a luxury. I'm offering you help with the thinking ahead part - hadn't even started any formal review yet, because the rationale behind the values of that change were missing and your subsequent editing showed you hadn't thought about it deeply. With a bit of back and forth, I am confident we can get something great out of it. If you don't have the time to discuss it or think through whether something is feasible or complete, you can always ask for help - on the PR.

> On smart fees - I definitely want to keep them away from the user [..] I am uncertain on how to resolve the issue with wallet providers.

If Dogecoin Core indirectly provides fee estimate data for users of 3rd party wallets (and specifically for that use case) then it should also - ideally - direct how to use that data properly; from a reference implementation point of view, but it may improve the quality of fees set by the wallet as a very welcome side-effect. However, as we will probably not be able to get a lot of field experience with it, perhaps the API endpoints can labeled as experimental, disclose that these are not fully field-tested yet, and try to establish a feedback loop, either through the issue system or through the discussions section here. It does need thorough testing though.

> [..] in particular having a difference in miner inclusion and minimum relay values

It's been like that since forever, until choices made - by you personally - for 1.14.2. The difference is that you're no longer in control of what shibes do, which is great, because 1.14.2 was... let's call it shenanigans, and not only have [legitimate concerns](https://www.reddit.com/r/dogecoin/comments/dtlkfp/dogecoin_1142_is_out_with_bugfixes/f7519fu/) been ignored, people that had [a differing opinion](https://github.com/dogecoin/dogecoin/commit/388fb79a3a78eae092c57e04c5a816697e14997d) have been driven away - even though they didn't have a totally invalid case. Thing is, developers can no longer control this coin's operations because there is insufficient competence. Instead, the community / market will have to figure it out. The trust previously offered by the community for you to manage it all, came with responsibility that was handled poorly, so it's gone. I've stived to dismantle threats to the coin - one of which being developer control - and will continue to ensure that developers, parasitic organizations and backchannel deals do not dictate the operation of this chain and coin. I don't care how many billionaires you bring on board or how many deals you think you can make: the times of developer control are over.

This means that, yes, now the software has to actually work without developer hardcoded controls. If that means that a block monitoring mechanism for fees is useful (I think it could be, if done right), then that's imho worth spending development time on. I think the main ideas from Bitcoin Core regarding how to do the monitoring are pretty solid, but we have to look at what gets monitored, what influences fee decisions and how. Nothing rocket-science there really, just assess, implement, test and maybe do that cyclical. I'll do a thorough review of the mechanics in your "smartfee" PR and suggest where things can get improved for a first iteration.

I also want to look deeper into RBF and CPFP mechanics, which, now that your relay prohibition is no longer in effect, become usable again.

> I'm using https://blockchair.com/dogecoin/nodes for node data, although that will be limited to nodes which are reachable

I don't know about yours, but all my nodes with < 250 connection slots are full, and for those, once they lose connection to blockchair's dummy node, there's a big chance that an actual node takes the slot. Only half of my nodes have a connection to it right now, because they are full. Doesn't mean these nodes don't exist though, and it means that enclaves are forming, also because the current developer-ran DNS seeds seem to be less optimized for large amounts of nodes on the network. I also see an odd 60-80 nodes popping on and off - mostly during US business hours - because someone spread the misinformation that it's very helpful to run a 1.14.4 relay node for 6 hours a day - and this makes the stats unreliable. 

So please, be cautious when you go out there and make decisions based on information if you don't have the ability to validate it yourself. In fact, could you please publish your plans here instead of doing them on proprietary channels, so that the community has a chance to react? Be open? Would be appreciated.

> We know F2Pool have upgraded, because we've talked to them about it

Great! I figured you took your contacts from hosting info@dogecoin.com with you into your new organization. I think I still have a list from @Sporklin somewhere with some more contacts. Let me know if you need help.

> and at least one other miner (because low fee transactions were being mined before F2Pool upgraded).

I'll try to get some monitoring going, per-block, to assess this more closely. However, since that apparent mystery miner did not have any blocks with low-fee tx within the set of 4000 I examined, I'm sure that you would agree that potentially waiting a couple of hours, or worst case, days, for all Dogecoin Core created transactions to be mined, to be a situation that no one benefits from. So can we please err on the side of caution rather than making rash decisions?

> If you're mostly happy with #2521, I'm happy to reopen

Please reopen and let's have the discussion. It's needed. If there is any problem with the parametrization I proposed, highlight it. If there's something I missed, add it. If you think it should be completely different, argue it. Disabling functionality is really not the answer, nor is doing shots in the dark.

> I'd appreciate thoughts on what estimatesmartfees should produce in 1.14.5

I'll spend time on #2564, but it probably won't be fully done before the weekend, because I have a queue of fixes for other things you flash-merged, which I need to make sure go in. You seem to be keeping the artificial release pressure up and I cannot trust your organization to have the discipline to refrain from pulling a release stunt with yet another broken software. I'll give you a preliminary review today or tomorrow though, without diving into every mechanic that now gets touched, so it won't be an ACK yet, but I hope to be able to contribute something meaningful to it.

rnicoll:
Closing, with #2567 replacing this.

PR #2577, opened 2021-09-19
wallet.h: change from bitcoin to dogecoin #ifndef

victorsk2019:
Proposed change for header guards naming from BITCOIN_WALLET... to DOGECOIN_WALLET... in this header file.  May change other header files, if necessary.

### DIFF
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -3,8 +3,8 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-#ifndef BITCOIN_WALLET_WALLET_H
-#define BITCOIN_WALLET_WALLET_H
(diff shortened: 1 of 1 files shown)


chromatic:
Looks like this PR reverts part of 9c6af6d84179e46002338bb5b9a69c6f2367c731 inadvertently. Can you rebase atop the most recent commit for 1.14.5-dev please?

chromatic:
I have no strong opinion about this change, but note:

 * there are almost 200 defines with the literal string `BITCOIN` in the 1.14.5-dev source code that could use a similar change
 * changing defines will make it slightly more difficult to cherry-pick, rebase, or merge changes from Bitcoin source that would otherwise apply cleanly

Is the 1.21 series more appropriate for such a change perhaps?

victorsk2019:
> Is the 1.21 series more appropriate for such a change perhaps?

Hi,

Yes, 1.21 is more appropriate.  Just brought it forward for future consideration.

PR #2588, opened 2021-09-24
Refresh English locale files

rnicoll:
This rebuilds the `bitcoinstrings.cpp` and `bitcoin_en.ts` files from the Dogecoin Core source, as well as correcting application names in all translation files. By correcting the application names Qt correctly recognises that translations are from the existing application, when refreshing other language files.


patricklodder (approved):
ACK. Great find on the DogecoinGUI -> BitcoinGUI fix. That makes sense.

PR #2601, opened 2021-10-02
Draft dust compatibility fix for 1.14.5

rnicoll:


patricklodder:
@rnicoll can you help me understand where in 1.14.4 source the `COIN` fee gets added during relay instead of `minRelayTxFeeRate`?

rnicoll:
> 
> 
> @rnicoll can you help me understand where in 1.14.4 source the `COIN` fee gets added during relay instead of `minRelayTxFeeRate`?

This comes from a bug report about the dust implementation, but I _think_ in the end the values in 1.14.4 are correct, just if we change the `IsDust()` threshold it needs to happen in a relay-only value first, then in the calculated value. Let me build up a reproduction case, this mostly exists because I had people asking about it, hadn't expected anyone to look at it yet.

rnicoll:
Oh also the bit where I wanted to take the calculation out entirely in this PR is just nonsense, I'll revert that shortly (it should be its own PR).

patricklodder:
> This comes from a bug report about the dust implementation

Can we see the report?

rnicoll:
dust

As said, wouldn't burn cycles on this until I have a recreation case for/against.

rnicoll:
Okay. Done a live relay test with:

* 1.14.4 miner
* 1.14.4 node in the middle
* 1.14.5 transaction creator

1.14.5 creates a transaction with a 0.1 Doge output and a <1 Doge fee, which 1.14.4 doesn't relay. Not sure why yet, though.

`010000000136acdab3b109c952aafc4b4c0c3a9f21be20bd23f2761ecf8b2cd7e9a3aac028000000006a473044022073b9a544bb246a710bb8e1cb165fdfaca7e252e700df94ad85f94f9b118f4e570220730d18a2c3286127049bc072792631720b239df97a72d41533acb4522aad60b20121039004b999717755276d213fa90f64679770781c8d1c85d49fb0130c9c03ac2212feffffff0280969800000000001976a9143404dba50cf4fa0046b9806d879a37631bf608fd88acb0d75905000000001976a914a0291cd23740f7f38de32e373e4fe4db401b99a888ac44593300` for the 1.14.5 generated transaction if you're curious

rnicoll:
1.14.4 version `v1.14.4.0-3ce34f6e6`, might be worth double checking that's the shipping version. Going to try adding debugging detail to 1.14.4 and go from there, because if we can't identify where it's dropping the transaction, can't fix it

rnicoll:
For my notes tomorrow; think 1.14.4 expects dust penalties if the output is <1 DOGE, and 1.14.5 only applies them well below that value. Transaction above has a fee of `0.00226` DOGE, I would expect `0.01226` DOGE I think (i.e. with a `0.01` DOGE penalty applied).

patricklodder:
>  I would expect 0.01226 DOGE I think (i.e. with a 0.01 DOGE penalty applied).

Besides that the tx you pasted has 2 dust outputs, not 1, the problem is:  

https://github.com/dogecoin/dogecoin/blob/4c93783ab64ba96e60b7ea37202c072c66f3f544/src/validation.cpp#L634

which triggers the dust rejection from:

https://github.com/dogecoin/dogecoin/blob/4c93783ab64ba96e60b7ea37202c072c66f3f544/src/policy/policy.cpp#L108-L110

It's not a matter of fee, but rather `-acceptnonstdtxn` / `ChainParams::RequireStandard()`, which is not enforced by default on regtest but it is on mainnet (so we have a misleading test issue here), which rejects the tx so the custom fee calculation from `GetDogecoinDustFee()` in `dogecoin-fees.cpp` is ignored.

PS: can we make this an issue report rather than a PR next time? ❤️

PR #2604, opened 2021-10-08
Add files via upload

69coin:
Update doge key


rnicoll:
I'm not sure this icon is actually displayed anywhere?

patricklodder:
@69coin do you have a screenshot with this icon? I also cannot seem to find where it's used. Just went through everywhere I thought it could be but couldn't find a single place where it's actually used.

alamshafil:
> I'm not sure this icon is actually displayed anywhere?

The icon is used here:
https://github.com/dogecoin/dogecoin/blob/3a29ba6d497cd1d0a32ecb039da0d35ea43c9c85/src/qt/bitcoingui.cpp#L375

Before:
Screenshot from 2022-07-21 14-41-44

After:
Screenshot from 2022-07-21 14-48-20

The key image becomes black & white since `TextColorIcon()` is being used.

patricklodder:
Thanks @alamshafil - that explains why I could not find it, because on a Mac those menu icons don't get rendered.

many missing

patricklodder:
Alright, I've thought about this and here's where I'm at:

1. This won't work with the current implementation, however
2. After we upgrade Qt to support dark mode, and
3. We find a better way to color the icons than by using `TextColorIcon`, this can still be a go.

Therefore I propose to keep this open, but mark it as blocked until 2 & 3 have happened. That way, we may ultimately get a cool icon, but in a way that will actually show a cool icon.

chromatic:
I did some research on this. It looks like the right way to do this is to create a custom style and use multiple icon variants, one for each style. That avoids the complex/expensive conversion work that `TextColorIcon` uses.

Bitcoin removed icons in menus altogether in [Bitcoin PR 16612](https://github.com/bitcoin/bitcoin/pull/16612) and `TextColorIcon` in [Bitcoin PR 16677](https://github.com/bitcoin/bitcoin/pull/16677), so anything we do here might conflict with 1.21.

I prefer keeping icons in the GUI despite the Bitcoin PR discussion, so I'm in favor of keeping this around until we have a stronger plan to address Patrick's points #2 and #3.

PR #2607, opened 2021-10-10
Update fundrawtransaction.py to match Bitcoin originals

rnicoll:
This revises the logic in fundrawtransaction.py to use the 20 small outputs generated, as inputs to the next transaction. This matches how the original worked, see https://github.com/bitcoin/bitcoin/blob/v0.14.3/qa/rpc-tests/fundrawtransaction.py#L556

### DIFF
--- a/qa/rpc-tests/fundrawtransaction.py
+++ b/qa/rpc-tests/fundrawtransaction.py
@@ -507,14 +507,14 @@ class RawTransactionsTest(BitcoinTestFramework):
         self.nodes[0].generate(1)
         self.sync_all()
 
-        for i in range(0,20):
+        for i in range(0,22):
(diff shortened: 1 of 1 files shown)


patricklodder (changes requested):
Tested against both 1.14.5-dev and #2606, this test fails on the latter.

When we raise the wallet's dust limit past 1 DOGE, as is needed in the immediate future, emptying the wallet at line 535 will cause there to not be enough left in the wallet to empty it. Suggest increasing the number of utxo generated on lines 510 and 540 by 2, so that there's always enough DOGE left to empty the wallet.

patricklodder (approved):
tested ACK on 1.14.5-dev and on top of #2606.

PR #2629, opened 2021-10-18
qa: fixes pruning test

xanimo:
Updated pruning test to reflect changes in ca3a26d

Enforce minimum block download window multiplier so test wouldn't timeout

### DIFF
--- a/qa/rpc-tests/pruning.py
+++ b/qa/rpc-tests/pruning.py
@@ -7,22 +7,76 @@
 # Test pruning code
 # ********
 # WARNING:
-# This test uses 4GB of disk space.
-# This test takes 30 mins or more (up to 2 hours)
+# This test uses 21GB of disk space.
+# This test takes 20 mins or more (up to 2 hours)
 # ********
-
+from test_framework.blocktools import create_coinbase
+from test_framework.mininode import CBlock
+from test_framework.script import (
+    CScript,
+    OP_NOP,
+    OP_RETURN,
+)
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import *
+from test_framework.util import (
+    start_node,
+    connect_nodes,
+    sync_blocks,
+    assert_equal,
+    assert_greater_than,
+    assert_raises_jsonrpc,
+)
 import time
 import os
-
-MIN_BLOCKS_TO_KEEP = 288
 
 # Rescans start at the earliest block up to 2 hours before a key timestamp, so
 # the manual prune RPC avoids pruning blocks in the same window to be
 # compatible with pruning based on key creation time.
 RESCAN_WINDOW = 2 * 60 * 60
 
+def mine_large_blocks(node, n):
+    # Make a large scriptPubKey for the coinbase transaction. This is OP_RETURN
+    # followed by 950k of OP_NOP. This would be non-standard in a non-coinbase
+    # transaction but is consensus valid.
+
+    # Set the nTime if this is the first time this function has been called.
+    # A static variable ensures that time is monotonicly increasing and is therefore
+    # different for each block created => blockhash is unique.
+    if "nTimes" not in mine_large_blocks.__dict__:
+        mine_large_blocks.nTime = 0
+        
+    # Get the block parameters for the first block
+    big_script = CScript([OP_RETURN] + [OP_NOP] * 950000)
+    best_block = node.getblock(node.getbestblockhash())
+    height = int(best_block["height"]) + 1
+    mine_large_blocks.nTime = max(mine_large_blocks.nTime, int(best_block["time"])) + 1
+    previousblockhash = int(best_block["hash"], 16)
+
+    for _ in range(n):
+        # Build the coinbase transaction (with large scriptPubKey)
+        coinbase_tx = create_coinbase(height)
+        coinbase_tx.vin[0].nSequence = 2 ** 32 - 1
+        coinbase_tx.vout[0].scriptPubKey = big_script
+        coinbase_tx.rehash()
+
+        # Build the block
+        block = CBlock()
+        block.nVersion = 0x620004
+        block.hashPrevBlock = previousblockhash
+        block.nTime = mine_large_blocks.nTime
+        block.nBits = int('207fffff', 16)
+        block.nNonce = 0
+        block.vtx = [coinbase_tx]
+        block.hashMerkleRoot = block.calc_merkle_root()
+        block.calc_sha256()
+        block.solve()
+
+        # Submit to the node
+        node.submitblock(block.serialize().hex())
+
+        previousblockhash = block.sha256
+        height += 1
+        mine_large_blocks.nTime += 1
 
 def calc_usage(blockdir):
     return sum(os.path.getsize(blockdir+f) for f in os.listdir(blockdir) if os.path.isfile(blockdir+f)) / (1024. * 1024.)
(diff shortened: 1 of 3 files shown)


xanimo:
Note: It was discovered during separating manual testing from the main test that pruneAfterHeight (1000) is lower than MIN_BLOCKS_TO_KEEP (1440) on testnet and regtest and this should be discussed/solved.

qa/rpc-tests/pruning.py
```diff
@@ -34,59 +93,56 @@ def __init__(self):
...
     def create_big_chain(self):
         # Start by creating some coinbases we can spend later
         self.nodes[1].generate(200)
         sync_blocks(self.nodes[0:2])
         self.nodes[0].generate(150)
-        # Then mine enough full blocks to create more than 550MiB of data
-        for i in range(645):
-            mine_large_block(self.nodes[0], self.utxo_cache_0)
+        # # Then mine enough full blocks to cre
+        # # ate more than 2200MiB of data
```
patricklodder, line 130: nit: strange looking comment

qa/rpc-tests/pruning.py
```diff
@@ -34,59 +93,56 @@ def __init__(self):
...
+        mine_large_blocks(self.nodes[0], 2395, self.prunedir)
 
-        sync_blocks(self.nodes[0:5])
+        sync_blocks(self.nodes[0:2])
+
+        mine_large_blocks(self.nodes[3], 995, self.prunedir)
+        mine_large_blocks(self.nodes[4], 995, self.prunedir)
```
patricklodder, line 136: I think this is what you meant by "Separated the manual testing from the main test"? So what we're saying is that it's faster / more reliable to mine 995 blocks than to sync 995 blocks? Even with the change to the download window? Doesn't have to be blocking for this PR, but I think this indicates further potential problems on the network stack.

qa/rpc-tests/pruning.py
```diff
@@ -10,19 +10,78 @@
...
+        node.submitblock(block.serialize().hex())
+
+        previousblockhash = block.sha256
+        height += 1
+        mine_large_blocks.nTime += 1
+
+        if len(dir) > 0:
+            if calc_usage(dir) > 2190 and calc_usage(dir) <= 2300:
+                print("current usage:", calc_usage(dir))
+                print(height)
```
rnicoll, line 84:
```suggestion
```

qa/rpc-tests/pruning.py
```diff
@@ -34,59 +93,56 @@ def __init__(self):
...
 
         # Determine default relay fee
         self.relayfee = self.nodes[0].getnetworkinfo()["relayfee"]
 
         connect_nodes(self.nodes[0], 1)
         connect_nodes(self.nodes[1], 2)
         connect_nodes(self.nodes[2], 0)
-        connect_nodes(self.nodes[0], 3)
-        connect_nodes(self.nodes[0], 4)
+        connect_nodes(self.nodes[3], 4)
```
rnicoll, line 121: Doesn't this mean nodes 3 & 4 are connected only to each other, not to node 0-2 & 5?

qa/rpc-tests/pruning.py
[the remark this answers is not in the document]
xanimo: Yes I wrote that so if you pass in self.prunedir to mine_large_blocks() you can debug what block heights are near our 2200 MiB MIN_DISK_SPACE_FOR_BLOCK_FILES. I can remove.

src/net_processing.cpp
```diff
@@ -3250,7 +3250,17 @@ bool SendMessages(CNode* pto, CConnman& connman, const std::atomic<bool>& interr
         if (state.vBlocksInFlight.size() > 0) {
             QueuedBlock &queuedBlock = state.vBlocksInFlight.front();
             int nOtherPeersWithValidatedDownloads = nPeersWithValidatedDownloads - (state.nBlocksInFlightValidHeaders > 0);
-            if (nNow > state.nDownloadingSince + consensusParams.nPowTargetSpacing * (BLOCK_DOWNLOAD_TIMEOUT_BASE + BLOCK_DOWNLOAD_TIMEOUT_PER_PEER * nOtherPeersWithValidatedDownloads)) {
+            int64_t nCalculatedDlWindow = 
+            std::max(consensusParams.nPowTargetSpacing, MIN_BLOCK_DOWNLOAD_MULTIPLIER) *
+             (BLOCK_DOWNLOAD_TIMEOUT_BASE + 
+             BLOCK_DOWNLOAD_TIMEOUT_PER_PEER * 
+             nOtherPeersWithValidatedDownloads);
```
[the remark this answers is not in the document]
xanimo: Sounds good I'll replace with your suggestion.

qa/rpc-tests/pruning.py
xanimo: Yes that's correct and why I'd like to discuss at some point what to do re nPruneAfterHeight (1000) being less than MIN_BLOCKS_TO_KEEP (1440) as mentioned above. I'll add a comment though.

src/net_processing.cpp
```diff
@@ -3250,7 +3250,17 @@ bool SendMessages(CNode* pto, CConnman& connman, const std::atomic<bool>& interr
...
+            if (nNow > state.nDownloadingSince + nCalculatedDlWindow) {
+                LogPrint("net", "Timeout downloading block: window=%d; inFlight=%d; validHeaders=%d; otherDlPeers=%d;",
+                 nCalculatedDlWindow, 
+                 state.vBlocksInFlight.size(), 
+                 state.nBlocksInFlightValidHeaders, 
+                 nOtherPeersWithValidatedDownloads);
                 LogPrintf("Timeout downloading block %s from peer=%d, disconnecting\n", queuedBlock.hash.ToString(), pto->id);
                 pto->fDisconnect = true;
```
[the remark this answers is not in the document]
xanimo: Yes it's intentional when providing -debug=net to gain deeper insight into the values of those 4 variables. The other message is not really helpful regarding debugging, at least to me in that it only provides hash and peer at timeout.

src/net_processing.h
```diff
@@ -21,6 +21,7 @@ static const unsigned int DEFAULT_BLOCK_RECONSTRUCTION_EXTRA_TXN = 100;
  *  Timeout = base + per_header * (expected number of headers) */
 static constexpr int64_t HEADERS_DOWNLOAD_TIMEOUT_BASE = 15 * 60 * 1000000; // 15 minutes
 static constexpr int64_t HEADERS_DOWNLOAD_TIMEOUT_PER_HEADER = 1000; // 1ms/header
+static const int64_t MIN_BLOCK_DOWNLOAD_MULTIPLIER = 10;
```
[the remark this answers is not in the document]
xanimo: > I think generally good to go, some nits to address then should be okay for final review. Thanks!

Sounds good, am working on these as I write. Thanks for the review!

src/net_processing.cpp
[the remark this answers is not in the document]
patricklodder: I requested this debug message so that when timeouts happen consistently, debug=net actually says something about this rather than just saying hey you timed out, as it's impossible to deduct this state from a node in the field. So the original `LogPrintf` line helps detecting issues when just running a node, the new `LogPrint("net", ...)` helps with actual debugging when things go south in the field.

src/net_processing.h
edtubbs, line 24: Consider writing a comment about multiplier relative to the test for future reference.

qa/rpc-tests/pruning.py
```diff
@@ -308,127 +361,127 @@ def has_block(index):
...
         connect_nodes(self.nodes[0], 5)
         nds = [self.nodes[0], self.nodes[5]]
         sync_blocks(nds, wait=5, timeout=300)
         self.stop_node(5) #stop and start to trigger rescan
-        start_node(5, self.options.tmpdir, ["-debug=1","-prune=550"])
+        start_node(5, self.options.tmpdir, ["-debug=1","-prune=2200"])
         print ("Success")
 
     def run_test(self):
         print("Warning! This test requires 4GB of disk space and takes over 30 mins (up to 2 hours)")
```
patricklodder, line 388: Here (and at the top level comment too) I think this should warn that this test takes (at its largest size before the bulk of all the pruning happened) a little under 21Gb, not 4.

rnicoll (approved):

patricklodder:
Hmm that sounds like something useful actually (it not being there) I'll go check bitcoin logs for it, see what happened to it.

xanimo:
https://github.com/dogecoin/dogecoin/blob/1.14.5-dev/src/chainparams.cpp#L440

PR #2674, opened 2021-11-06
Update build-unix.md

alfariqicrypto:
combine command. to make it easier when copying scrypt. remove the other sudo. only 1 is enough

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -64,8 +64,7 @@ using only fully tested dependencies, see the documentation in the
 
 **Required dependencies** :
 ```bash
-sudo apt-get install build-essential libtool autotools-dev automake pkg-config libssl-dev libevent-dev bsdmainutils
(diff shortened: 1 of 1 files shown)


AbcSxyZ (changes requested):
We may also add slashes to avoid long lines, which is also less convenient for copy pasting.

Do you think it could be great to also do the following ? :
```
sudo apt-get install build-essential libtool autotools-dev automake pkg-config libssl-dev libevent-dev \
    bsdmainutils libboost-system-dev libboost-filesystem-dev libboost-chrono-dev libboost-program-options-dev \
    libboost-test-dev libboost-thread-dev
```
It's possible to do the same for all code blocks where there is a scrolling bar, in `optionnaly dependecy` section & ARM-Cross compilation

alfariqicrypto:
o... its great... i love dogecoin

AbcSxyZ:
You could add the change rather than closing the PR :)

alfariqicrypto:
okay

Issue #2675, opened 2021-11-08
[bug] OpenType Font warning on Linux when first viewing Settings -> Options


chromatic:
# Bug Report

**Current behavior:**

On Linux, in the Qt interface, the first time I view the Settings -> Options menu, the GUI prints four lines to the `debug.log`:

```
GUI:   OpenType support missing for script 11
GUI:   OpenType support missing for script 11
GUI:   OpenType support missing for script 16
GUI:   OpenType support missing for script 16
```

The Options window eventually loads, and I can select the interface language from a dropdown list widget. Language and country names do appear to display correctly.

**Expected behavior**

The Options window displays correctly, including interface languages in a dropdown list. The long contains no warnings.

**Steps to reproduce:**

Launch the Qt application. When it loads, `tail` the `debug.log`. Then select Settings -> Options from the menu.

The first time this occurs, four error lines will appear in the log.

**Dogecoin Core version**

Reproduced in 1.14.3, 1.14.4, a pre-release of 1.14.5 generated by @rnicoll , and 1.14.5-dev compiled locally.

PR #2676, opened 2021-11-08
[Qt] Reduce font warnings

chromatic:
Addresses #2675 

Two points seem worth discussion:

- is the choice of Noto Sans reasonable
- should the font be loaded/initialized elsewhere, possibly included as a Qt resource


slightlyskepticalpotat:
It doesn't look like Noto Sans is [included](https://docs.microsoft.com/en-us/typography/fonts/windows_11_font_list) in Windows 11 or any other Windows versions, so we may want to bundle it. It's included by default with both macOS Catalina and Ubuntu 20.04.

patricklodder:
Missed that you targeted 1.21 here - reverting my tags

patricklodder:
Categorized this as "blocked" until we're at least somewhere after [`95cea3e` from #2304](https://github.com/dogecoin/dogecoin/pull/2304/commits/95cea3e8f1da72e280ea8d075577c4fc46b6df7e) in the 1.21 process.

chromatic:
I think this would work with 1.14.7. I don't have a strong preference, though reducing warnings pays off pretty quickly. This is still blocked on bundling that font file as a resource though.

chromatic:
I believe neither Comic Sans nor Comic Neue supports the glyphs necessary for the options dialog; that's why the warnings appear twice in #2675. Qt tries to use both fonts before falling back to a system font with the available glyphs.

chromatic:
I've bundled `Noto Sans` (adding `Roboto` was a typo; the font we want is called `[Noto Sans](https://fonts.google.com/noto/specimen/Noto+Sans/about)` licensed under the [Open Font License](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL)).

I've also rebased this on 1.14.7. Use the reproduction instructions in #2675 to test that this addresses the warnings.

slightlyskepticalpotat:
Going to build and try to confirm fix

slightlyskepticalpotat:
Just built on Ubuntu 22.04 LTS and tried to reproduce with instructions [here].

Option window loads quickly and displays correctly, including list of languages.

(https://github.com/dogecoin/dogecoin/issues/2675). 
image

image

Seems fully fixed to me!

patricklodder (approved):
ACK - tested on arm64 MacOS Ventura and x86_64 Ubuntu Trusty (for max backward compatibility)

@slightlyskepticalpotat tested on Windows

The only thing that's left is to comply with the OFL by linking it - we can pick that up as part of the "update copyright notices" task for the 1.14.7 release: I've added that to release plan.

PR #2765, opened 2021-12-19
Fix Char as Bool in interfaces

rnicoll:
Backport of https://github.com/bitcoin/bitcoin/pull/16572/commits/2dbfb37b407ed23b517f507d78fb77334142dce5

While `fSpent` is renamed in later versions, here it's under its original name.


fdoving:
utACK 06edc23f7409160adaaea5dd8d80b5dcaf696f99

patricklodder (approved):
tested ACK. Tested by copying a x86_64 wallet on linux from 1.14.4 to a new dogecoind both with and without `-rescan`.

PR #2795, opened 2022-01-10
Spelling error

anipaul2:


anipaul2:
> Hi, this error is already fixed by #2784. The main branch contain the latest release source code while all new changes are added to a development branch, [1.14.6-dev](https://github.com/dogecoin/dogecoin/tree/1.14.6-dev) currently.

Okay

ReverseControl:
Thanks @anipaul2. Already fixed as mentioned, but please keep submitting improvements like this. They are tiny things so we have a hard time tracking them down. Also, the `L` should be lower case in this instance.

anipaul2:
> Thanks @anipaul2. Already fixed as mentioned, but please keep submitting improvements like this. They are tiny things so we have a hard time tracking them down. Also, the `L` should be lower case in this instance.

ofc @ReverseControl. thanks for commenting!

PR #2923, opened 2022-04-14
Qt: Add prune setting to OptionsDialog

alamshafil:
This adds a prune setting to the options dialog. Backport from [Bitcoin PR #13043](https://github.com/bitcoin/bitcoin/pull/13043)

Screenshot from 2022-04-14 18-15-21

### DIFF
--- a/src/qt/guiconstants.h
+++ b/src/qt/guiconstants.h
@@ -5,6 +5,8 @@
 #ifndef BITCOIN_QT_GUICONSTANTS_H
 #define BITCOIN_QT_GUICONSTANTS_H
 
+#include "validation.h"
+
 /* Milliseconds between model updates */
 static const int MODEL_UPDATE_DELAY = 250;
 
@@ -52,5 +54,21 @@ static const int MAX_URI_LENGTH = 255;
 #define QAPP_ORG_DOMAIN "dogecoin.org"
 #define QAPP_APP_NAME_DEFAULT "Dogecoin-Qt"
 #define QAPP_APP_NAME_TESTNET "Dogecoin-Qt-testnet"
+
+/* One gigabyte (GB) in bytes */
+static constexpr uint64_t GB_BYTES{1000000000};
+
+/**
+ * Convert configured prune target bytes to displayed GB. Round up to avoid underestimating max disk usage.
+ */
+constexpr inline int PruneBytestoGB(uint64_t bytes) { return (bytes + GB_BYTES - 1) / GB_BYTES; }
+
+/**
+ * Convert displayed prune target GB to configured MiB. Round down so roundtrip GB -> MiB -> GB conversion is stable.
+ */
+constexpr inline int64_t PruneGBtoMiB(int gb) { return gb * GB_BYTES / 1024 / 1024; }
+
+// Default prune target displayed in GUI.
+static constexpr int DEFAULT_PRUNE_TARGET_GB{PruneBytestoGB(MIN_DISK_SPACE_FOR_BLOCK_FILES)};
 
 #endif // BITCOIN_QT_GUICONSTANTS_H
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -10,6 +10,7 @@
 #include "ui_optionsdialog.h"
 
 #include "bitcoinunits.h"
+#include "guiconstants.h"
 #include "guiutil.h"
 #include "optionsmodel.h"
 
@@ -41,8 +42,14 @@ OptionsDialog::OptionsDialog(QWidget *parent, bool enableWallet) :
     /* Main elements init */
     ui->databaseCache->setMinimum(nMinDbCache);
     ui->databaseCache->setMaximum(nMaxDbCache);
+    ui->pruneSize->setMinimum(DEFAULT_PRUNE_TARGET_GB);
     ui->threadsScriptVerif->setMinimum(-GetNumCores());
     ui->threadsScriptVerif->setMaximum(MAX_SCRIPTCHECK_THREADS);
+    ui->pruneWarning->setVisible(false);
+    ui->pruneWarning->setStyleSheet("QLabel { color: red; }");
+
+    ui->pruneSize->setEnabled(false);
+    connect(ui->prune, SIGNAL(toggled(bool)), ui->pruneSize, SLOT(setEnabled(bool)));
 
     /* Network elements init */
 #ifndef USE_UPNP
@@ -160,6 +167,9 @@ void OptionsDialog::setModel(OptionsModel *_model)
     /* warn when one of the following settings changes by user action (placed here so init via mapper doesn't trigger them) */
 
     /* Main */
+    connect(ui->prune, SIGNAL(clicked(bool)), this, SLOT(showRestartWarning()));
+    connect(ui->prune, SIGNAL(clicked(bool)), this, SLOT(togglePruneWarning(bool)));
+    connect(ui->pruneSize, SIGNAL(valueChanged(int)), this, SLOT(showRestartWarning()));
     connect(ui->databaseCache, SIGNAL(valueChanged(int)), this, SLOT(showRestartWarning()));
     connect(ui->threadsScriptVerif, SIGNAL(valueChanged(int)), this, SLOT(showRestartWarning()));
     /* Wallet */
@@ -179,6 +189,8 @@ void OptionsDialog::setMapper()
     mapper->addMapping(ui->bitcoinAtStartup, OptionsModel::StartAtStartup);
     mapper->addMapping(ui->threadsScriptVerif, OptionsModel::ThreadsScriptVerif);
     mapper->addMapping(ui->databaseCache, OptionsModel::DatabaseCache);
+    mapper->addMapping(ui->prune, OptionsModel::Prune);
+    mapper->addMapping(ui->pruneSize, OptionsModel::PruneSize);
 
     /* Wallet */
     mapper->addMapping(ui->spendZeroConfChange, OptionsModel::SpendZeroConfChange);
@@ -255,6 +267,11 @@ void OptionsDialog::on_hideTrayIcon_stateChanged(int fState)
     {
         ui->minimizeToTray->setEnabled(true);
     }
+}
+
+void OptionsDialog::togglePruneWarning(bool enabled)
+{
+    ui->pruneWarning->setVisible(!ui->pruneWarning->isVisible());
 }
 
 void OptionsDialog::showRestartWarning(bool fPersistent)
--- a/src/qt/optionsmodel.cpp
+++ b/src/qt/optionsmodel.cpp
@@ -9,6 +9,7 @@
 #include "optionsmodel.h"
 
 #include "bitcoinunits.h"
+#include "guiconstants.h"
 #include "guiutil.h"
 
 #include "amount.h"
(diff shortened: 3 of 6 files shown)


patricklodder:
Thanks! This is great!

That 2Gb in your screenshot is under our minimum threshold though, so I think we need to thoroughly test the min/default parameters to reflect this: https://github.com/dogecoin/dogecoin/blob/31afd133119dd2e15862d46530cb99424cf564b0/src/validation.h#L224-L225

alamshafil:
Should we hardcode the minimum threshold? Since the 2200MB is the minimum, we can use 3GB (rounded up).

alamshafil:
The code right now rounds 2200MB to 2GB as shown in the screenshot.

patricklodder:
> Since the 2200MB is the minimum, we can use 3GB (rounded up).

Yes, I'd recommend doing 3GB as a minimum (and default) (that's actually how I configure my own pruned nodes too)

src/qt/optionsdialog.cpp
```diff
@@ -41,8 +41,17 @@ OptionsDialog::OptionsDialog(QWidget *parent, bool enableWallet) :
     /* Main elements init */
     ui->databaseCache->setMinimum(nMinDbCache);
     ui->databaseCache->setMaximum(nMaxDbCache);
+    static const uint64_t GiB = 1024 * 1024 * 1024;
+    static const uint64_t nMinDiskSpace = MIN_DISK_SPACE_FOR_BLOCK_FILES / GiB +
+                          (MIN_DISK_SPACE_FOR_BLOCK_FILES % GiB) ? 1 : 0;
```
patricklodder, line 46: I would recommend moving this to `optionsmodel.h` (perhaps also as `constexpr`?) so that we can reuse it when setting a default (see next note)

src/qt/optionsmodel.cpp
```diff
@@ -90,6 +90,16 @@ void OptionsModel::Init(bool resetSettings)
     // by command-line and show this in the UI.
 
     // Main
+    if (!settings.contains("bPrune"))
+        settings.setValue("bPrune", false);
+    if (!settings.contains("nPruneSize"))
+        settings.setValue("nPruneSize", 2);
```
patricklodder, line 96: per the other note, if the minimum value `nMinDiskSpace` is moved to `optionsmodel.h`, we can then just use that here too.

patricklodder (commented):
Suggestion to change the default per our short discussion

src/qt/optionsdialog.cpp
alamshafil: We can remove the expression and just put a 3:
```
ui->pruneSize->setMinimum(3); // Set default to 3 GB (2200MB rounded)
```

src/qt/optionsmodel.cpp
alamshafil: We can use `3` and the code will set the prune size to `3000` MB.
Line 98 does this logic. (nPruneSizeMB)

src/qt/optionsdialog.cpp
patricklodder: That would mean that if we would change the value in `validation.h`, we need to *remember* that we need to change this value in multiple places (here and `optionsmodel.cpp`) too, but with my suggestion, it would automatically update it and we can just chill. 

That allows us to be lazy long-term by doing a little more now 😁

alamshafil: That makes sense :+1:

chromatic:
If we're using `1024 * 1024` instead of `1000 * 1000`, we should use `GiB` in user-visible text. That goes for `pruneSizeUnitLabel` in the Qt layout. As well, the calculation of `nPruneSizeMB` is correct, but I'd prefer we use one base unit consistently here across the GUI and the code.

The `--help` text suggests `MiB`, so my suggestion is to change the Qt label to `GiB` and the variable to `nPruneSizeMiB`.

alamshafil:
Bitcoin uses a constexpr in `guiconstants.h` so we can do the same. (3GB instead of 2GB)

patricklodder:
> If we're using 1024 * 1024 instead of 1000 * 1000, we should use GiB in user-visible text.

and

> Bitcoin uses a constexpr in guiconstants.h so we can do the same. (3GB instead of 2GB)

I agree with what both of you say, in principle, but I have some more thoughts about this.

The code in question (on Bitcoin Core's `master` branch) is:

https://github.com/bitcoin/bitcoin/blob/e14f0fa6a346afecbb1d5470aef5226a8cc33e57/src/qt/guiconstants.h#L56-L60
```cpp
/* One gigabyte (GB) in bytes */
static constexpr uint64_t GB_BYTES{1000000000};

// Default prune target displayed in GUI.
static constexpr int DEFAULT_PRUNE_TARGET_GB{2};
```

This leads me to think that:

1. **Bitcoin switched from GiB to GB.** I think that makes sense, because `dogecoin-qt` does most stuff in MB (but `dogecoind` does most in MiB) so it would be more consistent.
2. The *minimum* for Bitcoin using the provided calculation ( `MIN_DISK_SPACE_FOR_BLOCK_FILES / GiB + abs(sign(MIN_DISK_SPACE_FOR_BLOCK_FILES % GiB))` ) would be `550MiB/1GiB + 1 = 1GiB`, so it's (somewhat) understandable that they change the *default* to 2, also because Bitcoin Core allows for relay when pruning whereas our 1.14 does not. For our case however, I think that `2200MiB/1GiB + 1 = 3GiB` is pretty good as a minimum ***AND*** a default  - but I am open to other ideas; I'm biased as I've been running it like that for a while on mainnet because I felt it simply made sense to do it like that, no science, just gut.)

alamshafil:
I created `DEFAULT_PRUNE_TARGET_GB` in `optionsmodel.h` using this expression: `MIN_DISK_SPACE_FOR_BLOCK_FILES/1GiB + 1 = 3GiB`.

As of right now the `nPruneSize` calculates in MiB and the `DEFAULT_PRUNE_TARGET_GB` is calculated using MiB. 
Could we convert the GB to MiB? This is what Bitcoin Core `master` branch using `PruneGBtoMiB()`

patricklodder:
> I created `DEFAULT_PRUNE_TARGET_GB` in `optionsmodel.h` using this expression: `MIN_DISK_SPACE_FOR_BLOCK_FILES/1GiB + 1 = 3GiB`.

I'm okay with the `+1` instead of the translation of the modulus, because even if we would consider to set the value in `validation.h` to a MB size of a nicely rounded 1000 (which is a pretty ugly thing to do when we're setting it in the internals as MiB), we'd probably still want a +1 as both the default and the minimum, so I think that's fine.

> As of right now the `nPruneSize` calculates in MiB and the `DEFAULT_PRUNE_TARGET_GB` is calculated using MiB. Could we convert the GB to MiB? This is what Bitcoin Core `master` branch using `PruneGBtoMiB()`

The [`PruneGBtoMiB` and `PruneMiBtoGB`](https://github.com/bitcoin/bitcoin/blob/2e01b6986099715afa40ed6464da4b321b630e9c/src/qt/optionsmodel.h#L22-L30) functions are imho very useful and I hope that using this would please @chromatic too, if we apply them correctly: `PruneMiBtoGB` on the `DEFAULT_PRUNE_TARGET_GB` declaration and `PruneGBtoMiB` when using it in `SoftSetArg`.

I do think we'll need to declare them not as `static inline` functions but as `constexpr inline` functions to properly work with the assignment, and perhaps we can put it all (both inline functions and the `GB_BYTES` and `DEFAULT_PRUNE_TARGET_GB` constants) in `guiconstants.h` for consistency with ourselves and the current Bitcoin Core implementation

chromatic:
I agree with Patrick; that approach sounds reasonable.

src/qt/guiconstants.h
```diff
@@ -53,4 +55,20 @@ static const int MAX_URI_LENGTH = 255;
...
+ */
+constexpr inline int PruneMiBtoGB(uint64_t mib) { return (mib + GB_BYTES - 1) / GB_BYTES; }
+
+/**
+ * Convert displayed prune target GB to configured MiB. Round down so roundtrip GB -> MiB -> GB conversion is stable.
+ */
+constexpr inline int64_t PruneGBtoMiB(int gb) { return gb * GB_BYTES / 1024 / 1024; }
+
+// Default prune target displayed in GUI.
+static constexpr int DEFAULT_PRUNE_TARGET_GB{PruneMiBtoGB(MIN_DISK_SPACE_FOR_BLOCK_FILES)};
```
patricklodder, line 72: I think you missed the +1 here, otherwise awesome ❤️

alamshafil: Fixed!

chromatic (commented):
utACK; this matches what I expect.

I can test later this weekend, unless someone gets to it before I do.

alamshafil:
I noticed a problem:
```cpp
PruneMiBtoGB(MIN_DISK_SPACE_FOR_BLOCK_FILES + 1)
```
returns `4`, causing the UI default to `4` GB.

I played around with `PruneMiBtoGB` and it returns `5` GB when `MIN_DISK_SPACE_FOR_BLOCK_FILES` equals 3200 MiB.

Is there something wrong with calcucation? It looks like it is rounding it up no matter the number. Should we remove the `+1` to get `3` GB?

chromatic:
Now that I look at it in more detail, I'm not sure I understand the code in `PruneMiBtoGB` as implemented here. I would expect multiplying `mib` by `1024 * 1024`, then dividing by `GB_BYTES` (which is 1,000,000,000). I don't see the multiplication I expect here (I see it's in `MIN_DISK_SPACE_FOR_BLOCK_FILES`, but that makes me think the units may be misleading.)

If `MIN_DISK_SPACE_FOR_BLOCK_FILES` is measured in MB, then it should be multiplied in `PruneMiBtoGB`, but if it's measured in bytes, then it shouldn't be multiplied, but that probably means `PruneMiBtoGB` is implemented inconsistently with its name.

I do see adding `GB_BYTES - 1` which, when divided, gives the desired rounding behavior. With that in place, I don't think `DEFAULT_PRUNE_TARGET_GB` needs to add another `1` to the return value; this seems like the source of the extra GB.

Is that clear at all, or am I confusing myself writing this?

patricklodder:
Agree with both points:

> that probably means PruneMiBtoGB is implemented inconsistently with its name.

The units stated in the function name are misleading because the function does BytesToGB - good catch. For consistency we may want to rename it (because we should not want it to confuse us in the future)

> With that in place, I don't think DEFAULT_PRUNE_TARGET_GB needs to add another 1 to the return value; this seems like the source of the extra GB.

You're correct: I missed the rounding when I did my review, sorry for that.

alamshafil:
Looking at the `PruneMiBtoGB` implementation, it indeed does `BytesToGB`. Looks like the `1024 * 1024` is already done in `MIN_DISK_SPACE_FOR_BLOCK_FILES`

I have updated the function name to `PruneBytesToGB`.

> You're correct: I missed the rounding when I did my review, sorry for that.

No problem! I will fix this in the code.

chromatic (approved):
ACK.

I tested locally on 64-bit Linux, and it did what I expected.

patricklodder (approved):
ACK. Tested on a Mac M1 (GUI without wallet) and this works in both ways as advertised:

- after setting pruning and restarting the UI, pruning is applied
- after unsetting pruning and restarting the UI, the blockchain is reset and redownloads.

georgeartem:
was hoping to clarify, what am I passing for "n" if I am using `dogecoin.conf`

if it is `prune=1` how am I setting the size for what I want to prune to?

the command line help dialog says `prune=n` is incompatible if I have txindex=1 is that accurate? if` txindex=1 `is a requirement for my use case (I am using gigawallet) do you have any suggestions on how to set the pruning?

patricklodder:
> if `txindex=1` is a requirement for my use case (I am using gigawallet) do you have any suggestions on how to set the pruning?

You cannot both prune and have txindex activated because the transaction index references the file and position of the transaction on disk in the block file. So if you need txindex, you'll have to keep every historical block on disk.

Any other solution where we would store each transaction separately will have the same disk space requirement or worse, because `-txindex` is a global index.

georgeartem:
Thank you for clarifying.

PR #3000, opened 2022-06-19
Force `verifychain` RPC args to be positive

chromatic:
Verification levels must be between 0 and 4 inclusive, and block heights should always be positive. While code in the scan process handles the latter case with a default value, it's better to verify and reject invalid input where it first enters the system.


patricklodder (approved):
tested ACK (on Ubuntu focal)

PR #3015, opened 2022-06-25
Improve data types for network data

chromatic:
While working with other networking code, Patrick and I noticed that we use signed types and wrongly-sized types for several networking related values, including data lengths, timeout durations, and port numbers.

This commit corrects several of these types and improves error handling slightly to account for potentially invalid values.


patricklodder (approved):
ACK - tested on Ubuntu focal x86_64

PR #3050, opened 2022-07-17
update xanimo-key.pgp

xanimo:
just expired on thursday :D


patricklodder (approved):
ACK - key identifier remains the same

PR #2726, opened 2021-11-29
snap packaging for 1.14.6

slightlyskepticalpotat:
Please do not merge this yet, I will leave this as a draft until later in the release cycle in case any dependencies change.

Changes:
- Increment version number
- Make i386 and arm64 systems consistently build
- Add the tx and test binaries
- Update dependencies

PR #3107, opened 2022-08-27
Linux Snap Packaging for v1.14.6

slightlyskepticalpotat:
Got bored late yesterday and decided to resume working on https://github.com/dogecoin/dogecoin/pull/2726, which is much easier now that the new release has come out—I would love it if we could continue to work on decoupling this from the release cycle.

Changes:
- Extra testing for i386 and arm64
- Adds the tx and test binaries
- Uses lzo to speed up launch times
- Uses qt5 instead of qt4 to launch
- Simplifies installation process

I've built it successfully with Ubuntu 22.04 LTS (amd64) and Multipass, and launchpad's automatic build boxes have also built it on arm64 and i386 systems, though those were using 16.04. I've also unofficially uploaded it to the Snap Store, so anyone on a Linux box with Snap can test it with `sudo snap install dogecoin-unofficial`.

### DIFF
--- a/contrib/snap/README.md
+++ b/contrib/snap/README.md
@@ -1,24 +1,37 @@
-All the commands I used for building and pushing to Snapcraft. I've only tested the amd64 .snap package, but anyone on amd64, arm64, or i686 should be able to build and/or install it for themselves with these instructions.
+# Dogecoin Snap Packaging
 
----
+Commands for building and uploading a Dogecoin Core Snap to the Snap Store. Anyone on amd64 (x86_64), arm64 (aarch64), or i386 (i686) should be able to build it themselves with these instructions. This would pull the official Dogecoin binaries from the releases page, verify them, and install them on a user's machine.
 
 ## Building Locally
-
 ```
 sudo apt install snapd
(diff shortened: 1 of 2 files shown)


chromatic:
ACK; I built `dogecoin-core_1.14.6_amd64.snap` with this. My SHA-256 checksum is `6e0c64b4e2a82a33cb3d8386868afc3dbf3baba943c185d3c29e69eac1d58a9f`. If you can confirm this @slightlyskepticalpotat then we can merge.

slightlyskepticalpotat:
Unfortunately, snap builds are non-deterministic. However, it's definitely fine if it built and didn't error out or fail dogecoin tests during the process.

chromatic:
Ah, I hadn't realized. In that case let's tag @patricklodder ; do we want to move this all into another repository altogether now? I'm comfortable either way. This code is fine to merge into 1.14.7-dev, or it's fine to merge into another repo.

CryptoUnchained (approved):

patricklodder (approved):
utACK

Let's merge this here, and then we have a working script to extract.

PR #3138, opened 2022-10-16
Optimize IsHex() string helper function

chromatic:
This looks like a silly optimization, but this function gets called for transaction processing, so returning early when there's no point in even starting to iterate through a string is just smart coding.

### DIFF
--- a/src/utilstrencodings.cpp
+++ b/src/utilstrencodings.cpp
@@ -58,12 +58,16 @@ signed char HexDigit(char c)
 
 bool IsHex(const string& str)
 {
+    if ((str.size() <= 0) || (str.size()%2 != 0))
+        return false;
+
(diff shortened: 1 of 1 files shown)


patricklodder:
The only reason I can think of to do that check at the end is if you need the function to be operating in constant time. This function isn't doing that regardless because it breaks the loop on finding an invalid character.

I think that as long as we don't use this in p2p protocol or cryptographic operations (where we honestly should not be using hex strings anyway and I don't think we do) we should be fine - I'll iterate upwards through the callers one more time to make sure.

patricklodder:
Decided to invest time in setting up doxygen and graphviz.

image

Usages are:

- cli tools
- rpc methods
- rest endpoints

Optimizing this is fine and beneficial I think.

patricklodder (approved):
utACK

PR #3143, opened 2022-10-17
depends: tar: Always extract as yourself

xanimo:
For normal users, --no-same-owner is default, but not so for root, where it is assumed that root can change ownership willy-nilly. This is not the case for privilege-limited container environments where we gaslight the process into thinking it's root.

### DIFF
--- a/depends/funcs.mk
+++ b/depends/funcs.mk
@@ -78,8 +78,9 @@ $(1)_download_path_fixed=$(subst :,\:,$$($(1)_download_path))
 
 
 #default commands
+# The default behavior for tar will try to set ownership when running as uid 0 and may not succeed, --no-same-owner disables this behavior
 $(1)_fetch_cmds ?= $(call fetch_file,$(1),$(subst \:,:,$$($(1)_download_path_fixed)),$$($(1)_download_file),$($(1)_file_name),$($(1)_sha256_hash))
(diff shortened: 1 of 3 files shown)


xanimo:
> For normal users, --no-same-owner is default, but not so for root, where it is assumed that root can change ownership willy-nilly. This is not the case for privilege-limited container environments where we gaslight the process into thinking it's root.

For the record the above comment is from the original commit I cherry-picked this from (not my words :smile:). This is part of #3133 but separated for easier review.

patricklodder (commented):
Tested that this works with non-linux tar too (macOS) and checked the FreeBSD man page if this is supported and it is.

I have one tiny request: could you please add the short hash of the original **commit** that you cherry-picked to the commit message? I normally do:

```
Cherry-picked from: [commit]
```

for commits from bitcoin/bitcoin or other Dogecoin Core branches, and cite the source repo if it's coming from somewhere else. That way it's much easier to check the diff between the original commit and the picked result.

xanimo:
Absolutely! I will amend the commit with the original short hash when I get back to my workstation shortly.

xanimo:
Done :+1:

patricklodder (approved):
ACK - tested outside of linux CI with bsdtar 3.5.1

PR #3145, opened 2022-10-17
ci: drop extraneous libtinfo5 from x86_64-apple-darwin11

xanimo:
-drop duplicate of libtinfo5 in `x86_64-apple-darwin` specific continuous integration packages


patricklodder (approved):
utACK - does what it says

PR #3162, opened 2022-11-06
build: follow up to experimental feature guard

xanimo:
Have a few things to do re cleaning this up but it appears that the `HAVE_CONFIG_H` flag in support/experimentation.h was not being properly defined. This caused a cascade effect as the experimental compiler flags were then not configured/defined which I suspect were then not allowing the code to be compiled on the correct conditional path e.g. checking for `#ifndef USE_AVX2` wrapping `Round`, `Sigma`s, etc, (used) the `shaX_one_block_avx2` function (skipped) and the legacy SHA one block transformation (used) in all the sha routines.

This is in part based on the errors I encountered [here](https://github.com/xanimo/dogecoin/actions/runs/3398326504/jobs/5651272053) and it's corresponding experimental windows run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ build-aux/test-driver
 config.log
 config.status
 configure
+configure~
 libtool
 src/config/bitcoin-config.h
 src/config/bitcoin-config.h.in
--- a/configure.ac
+++ b/configure.ac
@@ -203,47 +203,48 @@ if test x$use_scrypt_sse2 = xyes; then
 fi
 
 AC_ARG_WITH([intel-avx2],
-  [AS_HELP_STRING([--enable-intel-avx2],
+  [AS_HELP_STRING([--with-intel-avx2],
   [Build with intel avx2 (default is no)])],
-  [use_intel_avx2=$enableval],
-  [use_intel_avx2=no])
+  [intel_avx2=$withval],
+  [intel_avx2=no])
 
-if test x$use_intel_avx2 = xyes; then
+if test x$intel_avx2 = xyes; then
   DOGECOIN_REQUIRE_EXPERIMENTAL
+  AC_MSG_CHECKING([whether to build with intel avx2 crypto])
+  AC_MSG_RESULT(yes)
+  AC_DEFINE(INTEL_AVX2, 1, [Define this symbol if intel axv2 works])
   case $host in
   *x86_64-*-linux*)
-      AC_CHECK_LIB([IPSec_MB],[sha1_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
-      AC_CHECK_LIB([IPSec_MB],[sha256_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
-      AC_CHECK_LIB([IPSec_MB],[sha512_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
-      AC_DEFINE(USE_INTEL_AVX2, 1, [Define this symbol if intel axv2 works])
-     ;;
+    AC_CHECK_LIB([IPSec_MB],[sha1_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
+    AC_CHECK_LIB([IPSec_MB],[sha256_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
+    AC_CHECK_LIB([IPSec_MB],[sha512_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
+    ;;
   *mingw*)
-      AC_CHECK_LIB([IPSec_MB],[sha1_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
-      AC_CHECK_LIB([IPSec_MB],[sha256_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
-      AC_CHECK_LIB([IPSec_MB],[sha512_one_block_avx2],LIBS=-lIPSec_MB, AC_MSG_ERROR(IPSec_MB missing))
-      AC_CHECK_LIB([Qt5PlatformSupport],[main],LIBS+=" -lQt5PlatformSupport", AC_MSG_ERROR(lib missing))
-      AC_CHECK_LIB([mingwthrd],         [main],LIBS+=" -lmingwthrd", AC_MSG_ERROR(lib missing))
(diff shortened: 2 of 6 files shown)


patricklodder:
🤔 The branch for the actions output you link [has over 10k changes, including adding custom SSE3 code](https://github.com/dogecoin/dogecoin/compare/1.14.7-dev...xanimo:dogecoin:1.14.7-dev-torv3-70-13191), so if I get this right, you're saying that when you are adding SSE3 code from upstream, we'll need to change the way the experimental AVX2 code currently works?

I don't however see any adaptation of that SSE3 code into our experimental framework, what are we trying to achieve?

xanimo:
I guess I should've opened an issue first but as is when compiling experimental for x86_64-linux-gnu running on 22.04 and with the same options as in CI, `USE_AVX2` is not being defined properly.

patricklodder:
So you're saying the code from 1.14.7-dev, as-is, is not linking intel-ipsec-mb when using `--enable-experimental` and `--with-intel-avx2`?

patricklodder:
I just ran the following against 4c31ab12 (=current tip of 1.14.7-dev), on 22.04:

```console
$ sudo apt-get install -y nasm
[..]
$ git clone git clone https://github.com/intel/intel-ipsec-mb.git
[..]
$ cd intel-ipsec-mb && git checkout v1.2 && make && sudo make install
[..]
$ ./autogen.sh && ./configure
[..]
  experimental  = yes
    SSE2 Scrypt   = no
    AVX2 crypto   = yes
    ARMv8 crypto  = no
    ARMv82 crypto = no
[..]
$ grep -1 AVX2 src/config/bitcoin-config.h
/* Define this symbol if intel axv2 works */
#define USE_AVX2 1

```

So I need your help to reproduce. Is this only for windows/depends on 22.04?

xanimo:
Hmm now I'm very confused. I just ran 2 separate builds starting with `./configure --prefix=`pwd`/depends/x86_64-linux-gnu --enable-experimental` and appending `--with-intel-avx2` and `--enable-scrypt-sse2` separately after having built depends with `AVX2=1` and then I just ran a 3rd time with both and all 3 tests passed.

This is what I had encountered just prior to that. :confused: 
```
/usr/bin/ld: libdogecoin_server.a(libdogecoin_server_a-init.o): in function `AppInitMain(boost::thread_group&, CScheduler&)':
init.cpp:(.text+0x10141): undefined reference to `scrypt_detect_sse2()'
/usr/bin/ld: bench/bench_dogecoin-scrypt.o: in function `Scrypt(benchmark::State&)':
scrypt.cpp:(.text+0x51): undefined reference to `scrypt_detect_sse2()'
/usr/bin/ld: libdogecoin_server.a(libdogecoin_server_a-init.o): in function `AppInitMain(boost::thread_group&, CScheduler&)':
init.cpp:(.text+0x10141): undefined reference to `scrypt_detect_sse2()'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:4067: dogecoind] Error 1
make[2]: *** Waiting for unfinished jobs....
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:4055: bench/bench_dogecoin] Error 1
/usr/bin/ld: test/test_dogecoin-scrypt_tests.o: in function `scrypt_tests::scrypt_hashtest::test_method()':
scrypt_tests.cpp:(.text+0xd6): undefined reference to `scrypt_detect_sse2()'
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:4257: test/test_dogecoin] Error 1
make[2]: Leaving directory '/mnt/source/repos/dogecoin/src'
make[1]: *** [Makefile:9852: check-recursive] Error 1
make[1]: Leaving directory '/mnt/source/repos/dogecoin/src'
make: *** [Makefile:694: check-recursive] Error 1
```

The full `config-opts` I was defaulting with earlier was: `./configure --prefix=`pwd`/depends/x86_64-linux-gnu --enable-experimental --enable-scrypt-sse2 --with-intel-avx2 --with-gui=qt5 --enable-zmq --enable-glibc-back-compat --enable-reduce-exports` which is the exact same as CI with exception to `--with-gui=qt5`.

patricklodder:
🤔 did you do this on a 100% clean dir? i.e. `make distclean` before `./autogen.sh`? There is some makefile magic to include [`crypto/scrypt-sse2.cpp`](https://github.com/dogecoin/dogecoin/blob/master/src/crypto/scrypt-sse2.cpp) and this sometimes causes grief - I do think we should fix that before lifting scrypt-sse2 out of experimental fwiw.

xanimo:
Did a `make clean` and just reconfigured. Did the same grep cmd and we can clearly see USE_AVX2 1:
image

Built and passed tests:
image

But what concerns/confuses me is vscode showing USE_AVX2 as not defined by the shading difference between what's defined and not:
image
image
image

xanimo:
I will do another round in a clean directory now. :crossed_fingers:

patricklodder:
`make clean` removes compiled files, not configuration. When you change anything in makefiles or configure.ac, you'll probably want to run `make distclean`, which removes everything.

xanimo:
Just ran both this branch and 1.14.7-dev in completely clean, freshly cloned directories:
```
git checkout 1.14.7-dev && make -C depends HOST=x86_64-linux-gnu NO_QT=1 AVX2=1 && ./autogen.sh && ./configure --prefix=`pwd`/depends/x86_64-linux-gnu --enable-experimental --enable-scrypt-sse2 --with-intel-avx2 --enable-zmq --enable-glibc-back-compat --enable-reduce-exports && grep -1 AVX2 src/config/bitcoin-config.h && make check -j4
```
and
```
git checkout 1.14.7-dev-runtime-guard && make -C depends HOST=x86_64-linux-gnu NO_QT=1 AVX2=1 && ./autogen.sh && ./configure --prefix=`pwd`/depends/x86_64-linux-gnu --enable-experimental --enable-scrypt-sse2 --with-intel-avx2 --enable-zmq --enable-glibc-back-compat --enable-reduce-exports && grep -1 AVX2 src/config/bitcoin-config.h && make check -j4
```

Both passed and had `USE_AVX2` defined but if you take look at the syntax highlighting here in `1.14.7-dev`:
image
compared to this branch `1.14.7-dev-runtime-guard`:
image

`#ifndef USE_AVX2` is not skipping `Ch`, `Maj`, `Sigma0-1`, `sigma0-1` and `Round` when it should.

`1.14.7-dev` is not calling `sha256_one_block_avx2` when it should and is instead performing the legacy routine:
image
compared to this branch:
image

In the grand scheme of things, this is rather insignificant to `1.14.7-dev` as it's an experimental feature and operates the same without `intel-ipsec-mb` but it completely defeats the purpose of itself, having an experimental features guard and is a waste of CI resources if this isn't a isolated instance, my suspicions are correct and can be confirmed by someone else.

Will follow up more tomorrow. Time to snooze. :sleeping:

xanimo:
Resolving this as an isolated incident. I just performed several more tests while deleting the sections that should be omitted and both branches passed several times. Chalking this up as some build cache/vscode/human error. Thanks for your time though.

PR #3167, opened 2022-11-08
Correct DEFAULT_BLOCK_MAX_SIZE to increase network Capacity

microguy:
The maximum allowed size for a block in /src/consensus.h (protocol rule), excluding witness data, is 1 MB. The default setting for DEFAULT_BLOCK_MAX_SIZE should be 1 MB if we are trying to fully maximize the network.

Some pools might patch this on their own anyway. But setting this at 750K might lead to reduced network throughput.

### DIFF
--- a/src/policy/policy.h
+++ b/src/policy/policy.h
@@ -23,7 +23,7 @@ class CCoinsViewCache;
 static const CAmount RECOMMENDED_MIN_TX_FEE = COIN / 100;
 
 /** Default for -blockmaxsize, which controls the maximum size of block the mining code will create **/
-static const unsigned int DEFAULT_BLOCK_MAX_SIZE = 750000;
+static const unsigned int DEFAULT_BLOCK_MAX_SIZE = 1000000;
 /** Default for -blockprioritysize, maximum space for zero/low-fee transactions **/
 static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = 27000;
 /** Default for -blockmaxweight, which controls the range of block weights the mining code will create **/


GiverofMemory:
I would expect that our orphan rate is negligible currently, but perhaps setting an acceptable orphan rate % would be wise so we have a target when pursuing changes in the future.  I would think 1% would be fine.  In terms of orphans we should consider what are the negative consequences.  Some negatives I can think of is wasted work (1% wasted work seems negligible), and mining unfairness.  Again 1% unfairness benefit for large pools mining two blocks in succession isn't too horrible.

If you come up with a way to gather that data easily perhaps with libdogecoin or something let me know and I can help data gathering and analysis.

chromatic:
I have two thoughts on this.

First, it's nice to have consensus between the mining defaults and the consensus defaults, so if we don't make both the same, we should have a good comment and documentation as to why. Design ACK from me.

Second, I agree with measuring this... somehow. What would make us want to change this number and how would we know we should consider it? I'm not sure that we have to answer that before deciding how to handle this PR, but it's a good opportunity to raise the question. If that should become a discussion item instead, I'm happy to move this conversation out of the PR comments.

philipperemy:
@patricklodder Can you post the result of the analysis to see if we need to increase the block size and push the change? Thanks!

philipperemy:
Here's some data about the average block size: https://bitinfocharts.com/comparison/size-doge-dogecoin.html#alltime.
It has been a bit congested since December 2023. I guess we should have a discussion if we should increase it or not.

microguy:
> Here's some data about the average block size: https://bitinfocharts.com/comparison/size-doge-dogecoin.html#alltime. It has been a bit congested since December 2023. I guess we should have a discussion if we should increase it or not.

Yes. The Dogecoin network is being congested by the limitation of this arbitrary 750 KB cap.

doge_cong

Dogecoin has an advertised block size of 1 MB. But in reality, the network has just 75% of that capacity due to the conflict addressed and solved in this PR.

PR #3207, opened 2023-01-21
compatibility with qt 5.15.8 and boost 1.81.0

bjacquin:


chromatic:
utACK on the design; seems well worth testing.

chromatic:
I tested this with a `bookworm-slim` image. Code compiles fine, but I run into the same OpenSSL SHA making test failures we've seen before. I think that's probably okay but would appreciate another core maintainer looking this over.

Issue #3216, opened 2023-02-04
[feat] Promote HTTP RPC Binding Logs out of Debug Logs


chromatic:
# Feature Request

**Describe the Feature Request**

When the Core starts up and RPC traffic is allowed, the server binds to specific interfaces and ports. Configuration options allow further customization.

Data on this is only provided in the logs when `http` is in one of the allowed logging categories. This category also provides logging for HTTP connections for RPC, and may not be useful or desirable as a general logging category.

It's useful to know where and how the Core is listening for RPC connections, especially if you're trying to secure and keep a node secure.

**Describe Preferred Solution**

Remove the `http` category from `LogPrint` lines in `src/httpserver.cpp` where the Core logs information about binding to addresses and ports and any logs about including or excluding IP addresses and networks with netmask or CIDR information.

**Describe Alternatives**

By default, the core only logs failed binding information, which is helpful but not sufficient for auditing purposes. Users can get this information by enabling *all* `http` category logs, but that's a lot of additional information which may not be useful.

**Related Code**

See [IP allowed logging](https://github.com/dogecoin/dogecoin/blob/1.14.7-dev/src/httpserver.cpp#L222-L225) and [IP address binding logging](https://github.com/dogecoin/dogecoin/blob/1.14.7-dev/src/init.cpp#L1382-L1387), especially the inconsistency between failure and success modes in the latter.

PR #3217, opened 2023-02-05
Remove http log category from RPC binding-related logs

bwarden:
Implement #3216

### DIFF
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -222,7 +222,7 @@ static bool InitHTTPAllowList()
     std::string strAllowed;
     for (const CSubNet& subnet : rpc_allow_subnets)
         strAllowed += subnet.ToString() + " ";
-    LogPrint("http", "Allowing HTTP connections from: %s\n", strAllowed);
+    LogPrintf("Allowing HTTP connections from: %s\n", strAllowed);
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


src/httpserver.cpp
```diff
@@ -222,7 +222,7 @@ static bool InitHTTPAllowList()
     std::string strAllowed;
     for (const CSubNet& subnet : rpc_allow_subnets)
         strAllowed += subnet.ToString() + " ";
-    LogPrint("http", "Allowing HTTP connections from: %s\n", strAllowed);
+    LogPrint("Allowing HTTP connections from: %s\n", strAllowed);
```
chromatic, line 225: I think this and the other should be `LogPrintf` as the first argument is a `sprintf`-style string.

src/httpserver.cpp
```diff
@@ -345,7 +345,7 @@ static bool HTTPBindAddresses(struct evhttp* http)
 
     // Bind addresses
     for (std::vector<std::pair<std::string, uint16_t> >::iterator i = endpoints.begin(); i != endpoints.end(); ++i) {
-        LogPrint("http", "Binding RPC on address %s port %i\n", i->first, i->second);
+        LogPrint("Binding RPC on address %s port %i\n", i->first, i->second);
```
chromatic, line 348: See line 353 here; I think that approach will work. If you force push a commit, I'll enable CI and run the tests myself too.

chromatic:
ACK on 64-bit x86 Linux. In the `debug.log` now I see the lines:

```
2023-02-06 18:08:43 Binding RPC on address ::1 port 22555
2023-02-06 18:08:43 Binding RPC on address 127.0.0.1 port 22555
```

If you squash these two commits together and force push, this should be ready for final review. Thanks!

chromatic (approved):
Re-ACK. Good to go. Thank you and congratulations on your first contribution!

PR #3218, opened 2023-02-07
Add Dockerfile generation script

chromatic:
This script generates a Debian/Ubuntu-style Dockerfile that can checkout a PR and run tests locally. I don't want to expand it too much further in terms of setting up a development environment, but it's a base we can build on. I'm leaving this as a draft for now for further discussion (including "what are you doing?!").


chromatic:
I have some answers! Some!

1. Probably. I think this will be most valuable if we can select from a couple of different types of Linux machines, including a Fedora and a Debian/Ubuntu derivative.
2. I'd like to be able to take out the compiled binaries, but I haven't gotten that far yet.
3. I'm trying to make my life easier, and I figure if my life is easier, other people might want similar things. Mostly I'd like people to be able to run targeted tests in certain circumstances ("Let's see what we get with gcc 12!") without having to reinvent the wheel.

chromatic:
I'm not sure I understand the goal of 1 yet. I think we can defer the goal of 2. It's worthwhile, but it's not urgent.

For 3, that seems the most useful. I'd like to be able to point other people at this and say "Here are some options that do the repetitive but complex but repetitive 80% of the work for you and get it out of the way".

chromatic:
The images defined in our GH workflow files? I think the intent is similar but not the same. The results should be similar, but if I want to point a new developer at something, I'd rather say "Here's a way to build this PR on your own laptop and see what happens" without explaining how to get images out of CI or write their own `Dockerfile` or set up the right compilation environment.

I haven't tried doing the same with our CI images though, so if that's an easier path I'm happy to use that instead.

chromatic:
We might be talking past each other here. I'd like to be able to point an interested contributor at a PR and say "Here's the easy way to build this yourself, run tests, and potentially make changes and re-run the process" which is different from "Here are the artifacts built from this PR so you can carefully run manual regression tests". Both are valid, but they come at it from different angles.

philipperemy:
Some comments:

- I'd add `DEBIAN_FRONTEND=noninteractive`.
- Also probably not the best way to (default) Ubuntu version `ubuntu:jammy`.
- Does it work if the `pr` parameter is blank (unspecified)?

Issue #3152, opened 2022-10-24
[feat] Remove hard-coded paper size in printer configuration


chromatic:
# Feature Request

The current Dogecoin Core Qt printer setup has a hard-coded page size of A4 ([1.14.7-dev link to source code](https://github.com/dogecoin/dogecoin/blob/1.14.7-dev/src/qt/utilitydialog.cpp#L360). For people with other paper sizes (letter, in the US, for example), this may render printable items slightly off the page.

**Describe Preferred Solution**

The ideal solution is to let Qt auto-detect the preferred page size. Qt suggests this may be possible in the [QPrinter documentation](https://doc.qt.io/qt-5/qpagesize.html#details):

> A QPageSize instance may also be obtained by querying the supported page sizes for a print device. In this case the localized name returned is that defined by the printer itself. Note that the print device may not support the current default locale language.

**Describe Alternatives**

Alternately, we could prompt the user for a paper size.

I'm comfortable coding this myself, but I'd like to review the feature idea and suggested implementation with others before doing so. (I'm also happy to review a pull request if someone else wants to code it.)

jadijadi:
Can this be closed? It seems the PR #3239 have fixed it.

PR #3239, opened 2023-03-16
Make Qt auto-detect the preferred page size for printer

Anish-M-code:
-> Removes the need for hard-coded paper size in printer configuration if QT 5.3 or above is used related to issue #3152

### DIFF
--- a/src/qt/utilitydialog.cpp
+++ b/src/qt/utilitydialog.cpp
@@ -58,6 +58,7 @@
 #include <QtPrintSupport/QPrinter>
 #include <QtPrintSupport/QPrintDialog>
 #include <QtPrintSupport/QPrintPreviewDialog>
+#include <QtPrintSupport/QPrinterInfo>
 #endif
 #include <QPainter>
 #include "walletmodel.h"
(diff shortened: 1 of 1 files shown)


chromatic:
This looks generally correct to me; concept ACK.

I wonder if it would read better AND work correctly if there were only one `#define` block; is it possible to get and set paper/page size that way?

Anish-M-code:
should i enclose my changes in a single #define block  and like #define USE_DEFAULT_PAGE_SIZE and set it true to conditionally compile my existing changes.

chromatic:
I tested this and the code compiles and passes tests fine; I need to test again with a printer on my network (it just disappeared during testing).

Printing to PDF suggests a paper size of A4, but I'm not sure if/how this would address that issue. So far so good for the effects.

Anish-M-code:
@chromatic kindly let me know in case if i had to change code if code doesnt work as expected, i usually test code before submitting PR but i didnt have printer setup so i couldnt test this one.

Anish-M-code:
@chromatic kindly let me know if my feature doesn't work as expected, I usually test code before submitting PR but i didnt have printer setup so i couldnt test this one.

chromatic:
I apologize for the delay.

Tested ACK -- code builds and passes all tests. I tried running this code without an active printer on the network, and the paper size defaulted to A4. Then I connected a printer to the network with a different paper size (US Letter by default), and the print dialog used the printer's default.

I'd like to have the commits squashed before merging. If you're willing to do that @Anish-M-code, that'd be great. Otherwise one if the core maintainers can do it for you. Do you have a preference?

chromatic:
I've rebased and squashed, so this could use one more maintainer's review before a final merge.

patricklodder (approved):
ACK. Tested on Windows 11.

- Paper size selection ✅
- Aspect ratios retained ✅
- Default A4 fallback ✅

Apologies for this taking so long to confirm. I had to do this while visiting family that (a) had supports a printer with multiple paper sizes and (b) uses Windows PCs I could test on.

Issue #3201, opened 2023-01-11
[bug] Regression Testing Core Initial Build -extended


DogestMaximus:
Regression Testing Core Initial Build -extended

# Bug Report

**Current behavior:**
Total of 6 Fails in regression testing -extended initial build of Core

**Steps to reproduce:**
This was reproduced both in Debian 11 (KDE) and Debian 11 VM (XFCE)

**Screenshots**

![image](https://user-images.githubusercontent.com/119521071/211733493-f660c095-d743-41de-9e19-6595ea64e9e8.png)
![image](https://user-images.githubusercontent.com/119521071/211733651-ce7c76f0-775c-4666-a52b-0f89c5ce2388.png)

**Dogecoin Core version**

Dogecoin Core 1.14.6, official binary, self compiled, headless

**Machine specs**
- OS: Debian 11 (KDE) && Debian 11 VM (XFCE)
- CPU: Intel Core i5 && KVM x 4
- RAM: 12G && 8G
- Disk size: 1TB && 100GB
- Disk Type (HD/SDD): SSD && SSD

**Extra information**

smartfees.py
rpcbind_test.py
bip65-cltv.py
bipdersig.py
p2p-feefilter.py
maxuploadtarget.py

DogestMaximus:
Reproduced on both Laptop Intel Core i5 KDE Plasma sddm and Virtual Machine (ProxMox Debian) Intel Core i7 XFCE lightdm

chromatic:
I tested `maxuploadtarget.py` a couple of times. This has the feeling of a fencepost error. Whether it's in the disconnection code or the test, I'm not sure. 3874 times through the loop is slow to test though.

DogestMaximus:
> I've been looking at this in VSCode running on the VM with XForwarding attempting to step through the issue (no idea what I'm doing with it yet) but according to some "research" It seems that the all_interfaces() function is trying to call the tostring() method on an array.array object, but this method does not exist- alleging that the all_interfaces() function is trying to use an outdated version of the array module, which was removed in python 3.x.  If it's not an emergency issue I can have a better look at it this week once I become more familiar with both python --version, their nuances- and VSCode itself.  If someone else gets to it first that's also fine- just let me know or update this thread.

$ python3 qa/rpc-tests/rpcbind_test.py
Initializing test directory /tmp/testxtc5mr54/504772
Unexpected exception caught during testing: AttributeError("'array.array' object has no attribute 'tostring'")
[...]

PR #3241, opened 2023-03-26
Update fee rate to 0.01 DOGE/kb in p2p-feefilter.py

Anish-M-code:
I have made an attempt to update fee rate in p2p-feefilter.py to recommended minimum transaction fee of 0.01 DOGE/kb to reflect fee rate changes made in 1.14.4 release , it is linked with issue #3201


chromatic (approved):
ACK; tests pass and code reviews fine. 64-bit x86 Linux. Thank you!

PR #3248, opened 2023-04-10
Add getblockchainstats rpc command

chromatic:
I found myself wanting a command like this. I have no strong feelings about the names and I'm open to implementation improvement suggestions. Right now I'm most interested in feedback, especially of the kind "You should do X instead!"


chromatic:
Good pointer, thanks. I found an approach that appears to scale much better by avoiding disk IO, but I'm not sure if it has similar accuracy.

```
a_machine:~/dev/dogecoin$ time src/dogecoin-cli getblockchainstats 50000
{
  "height": 4670739,
  "bestblockhash": "94bd899a9f35926e9bb117793b9236cde2b77a54e3be820da8bb19ac847cde83",
  "transactions": 1115802,
  "count": 50000
}

real	0m0.982s
user	0m0.022s
sys	0m0.000s
a_machine:~/dev/dogecoin$ time src/dogecoin-cli getblockchainstats 50
{
  "height": 4670756,
  "bestblockhash": "3d631b059cec6af8471aff688fcd10c4b75f9a517ce9ddb9382e868530c7b950",
  "transactions": 871,
  "count": 50
}

real	0m2.798s
user	0m0.015s
sys	0m0.006s
a_machine:~/dev/dogecoin$ time src/dogecoin-cli getblockchainstats 500
{
  "height": 4670767,
  "bestblockhash": "0a16017c7442da60a5708088a600decc7f0120c08e91f83f132a7acec9af1f44",
  "transactions": 7359,
  "count": 500
}

real	0m3.692s
user	0m0.003s
sys	0m0.004s
a_machine:~/dev/dogecoin$ time src/dogecoin-cli getblockchainstats 1000
{
  "height": 4670789,
  "bestblockhash": "a2b0ab81e76f290ce9a7a1308be420b50d2691c0efb43854504fe993751c4b89",
  "transactions": 14328,
  "count": 1000
}

real	0m3.288s
user	0m0.015s
sys	0m0.006s
a_machine:~/dev/dogecoin$ time src/dogecoin-cli getblockchainstats 100000
{
  "height": 4670805,
  "bestblockhash": "9af9af5ace3e8dc9fce3d99bba5b17b501a866ae4e6fa9de356bb414affca967",
  "transactions": 2126021,
  "count": 100000
}

real	0m3.395s
user	0m0.020s
sys	0m0.001s
```

chromatic:
I've tried three approaches here, and think a hybrid of the second and third might be the way to go. The third is the fastest, though the comments for the `CBlockIndex` declaration suggest the value I use here might not always be available, so I could fall back to the second approach in that case.

The name of this command seems less useful now, so it may be worth bikeshedding that.

Alternately, if we have another command that gets block stats, would it be worth extending the output there? I'm less convinced, given the potential for fallback behavior.

Issue #3243, opened 2023-04-04
remote p2p bandwidth/cpu attack


visualbasic6:
hi,

there is a remote attack that can be carried out in perpetuity against vulnerable nodes in a way that a single attacking machine can force `dogecoind` to upload > 130MB/s to the attacking machine in a sustained way.

doge nodes on isps or cloud providers that charge for or throttle upstream after a cap has been reached are, unless hardened against this attack at a server level, vulnerable to the attack.

i did some testing and ran my digitalocean bandwidth costs to > $175 relatively quickly.

![image](https://user-images.githubusercontent.com/38997186/229753562-079d844f-8b82-4d23-9fd8-56fc6bcca3d7.png)

making this a financial attack against vulnerable nodes - but not an attack capable of taking the network offline.

the attack is carried out by an [evil multithreaded script](https://pastebin.com/raw/NFU9KM70) that pretends to be several valid nodes. this is accomplished by operating in the network at a low p2p protocol level. after performing a complete handshake with the target node, resulting in a healthy peer connection, the script begins to request mempool content in a loop. at this point a single attacking machine can trigger, again, `dogecoind` into sending > 130MB/s worth of mempool data. additionally - with code alterations it is likely that the evil script could target multiple dogecoin nodes simultaneously, from one machine, assuming that certain conditions are met on the attacker's end.

i can't find a bug bounty program so i'm just posting it here. it's not really the exploit of the century so i'm not concerned with evil actors carrying out bandwidth attacks against their enemies. that said - still - [get that fixed](https://youtube.com/watch?v=BU3tDES29sY&t=143s).

also - it is extremely likely that bitcoin is also vulnerable to the attack 🤭 

but blockstream (skids) and their approved programmers (skids) are ants at a picnic when it comes to irresponsible public disclosure. i'd like to think that, at least here in the doge community, we like to have a bit of fun and [do a little trolling](https://youtu.be/PobQzVsj7GE?t=4).

original src: https://twitter.com/123456/status/1643153544067837955

thug life,
pad

michilumin:
Ok. Eyes on this on my end as well..  In addition to tracking a limit of amount of data sent to a single requesting node, some heuristic trigger which could identify a bogus loop of requests versus legitimate ones.  Wouldn't want to implement a rate limit that slows the network down as initial synchronization is already a pain point for setting up a new node. On radar, though.

GiverofMemory:
[...]

Issue #3206, opened 2023-01-19
[feat]  Retain Lock Unspent  on Core restart


qlpqlp:
# Feature Request

**Describe the Feature Request**
I like that I can Lock unspent Wallets but it always unlock wen restarting my Core Wallet, and I have always to manually Lock again unspent Doge

**Describe Preferred Solution**
Retain in the configuration memory the locked unspent wallets

**Describe Alternatives**
Show a warning that if you locked any unspent wallets, they are now unlocked on restart

**Related Code**
![image](https://user-images.githubusercontent.com/5968076/213437928-8a61b32c-f668-4361-9785-a49551e5aee0.png)

**Additional Context**
Keep up the great work shibes :)

EvensD9:
HI
HOW I DO FOR PROGRES IN THE DOGECOIN

evil-user:
@qlpqlp  is this a beginner friendly feature request , i would like to contribute, would loading the configuration of locked unspent wallets in a configuration file and reading config file while starting core wallet satisfy this feature request?

qlpqlp:
Much thanks @patricklodder, it will help shibes, I think, I personally use it a lot but restarting is always dangerous because I have to manually lock again every unspent to not use it :)

@evil-use Much thanks also for replying :)

Skylerspencer:
But could it be that maybe you bought a used wallet that belong to somebody else and didn't know it  surely they could have taking the money are there such thing as lost and found and they are saying hey this was stolen from me I'm not your enemy but friend somebody tricked you because you I  thought crypto stays based the wallet stays with its owner  I have never seen used wallets for sell oh shit did I just  - [ ] discover-  a secret code shutting down be back later[ ] no matter were it goes that's one of the mistaken truth people don't know   not trying to cause an issue new at this trying to understand because that happened to me and it is still happening. I lost billions still am I wonder who really is that person that  has done this to us .

Skylerspencer:
I created  One wallet with 900,000,000 invested in it and now it has over 200 moe wallets going in and out of it when I got stolen who is it thats that smart to pull something this huge off  not only two years of high school  this is an issue that must be resolved. Or were all in trouble.

qlpqlp:
Hey @Skylerspencer  what you are saying has nothing to do what  I posted. If you lost Doge you should know that is your responsibility to make sure your own Wallet is always safe.

PR #3255, opened 2023-04-30
Add setmaxuploadtarget RPC command

chromatic:
Addresses part of GH #3243.

This needs tests if we're going to go much further, but I'm more interested in concept review. Does it make sense at all? As written? Are there any glaring errors that need design thought before going further?


chromatic:
Good point. We already have a request for something similar in #3206. I think this gives us enough desired use cases to design that feature.

PR #3257, opened 2023-05-08
Add height parameter to walletnotify message

chromatic:
Use %i in the command to include the height of the block containing the transaction.

I found myself wanting this when trying to perform an action on an incoming transaction while awaiting _n_ block confirmations. Otherwise I have to perform multiple RPC calls to get this information.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -33,6 +33,7 @@
 
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/filesystem.hpp>
+#include <boost/lexical_cast.hpp>
 #include <boost/thread.hpp>
 
 using namespace std;
@@ -997,7 +998,9 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFlushOnClose)
 
     if ( !strCmd.empty())
     {
+        int64_t nHeight = wtxIn.hashUnset() ? 0 : mapBlockIndex[wtxIn.hashBlock]->nHeight;
         boost::replace_all(strCmd, "%s", wtxIn.GetHash().GetHex());
+        boost::replace_all(strCmd, "%i", boost::lexical_cast<std::string>(nHeight));
         boost::thread t(runCommand, strCmd); // thread runs free
     }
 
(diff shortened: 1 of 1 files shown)


sinetek:
looks good to me

chromatic:
I think the existing mechanism already has the chain tip fork problem, so if we're going to solve that problem, we should solve it for block/wallet notification mechanisms in general.

I'll add the appropriate `-help` text.

chromatic:
Oh, I see what you mean. In that case we have a couple of options:

- add an additional parameter to disambiguate new from updated
- skip the height on updated transactions
- use a flag value (-1? 0?) for updated transactions

I tend to prefer unambiguous data, so my slight preference is the first option. All of these are easy enough to code and explain.

chromatic:
Something more like this you think?

`int64_t nHeight = wtxIn.hashUnset() ? 0 : mapBlockIndex[wtxIn.hashBlock]->nHeight;`

No extra parameter, no extra formatting token, present only when the user has asked for the height as a parameter, unambiguous and static flag value that I can document when I update `-help`?

chromatic:
Squashed and rebased.

patricklodder (approved):
tested ACK, but with a discovery:

When testing, I found that when reversing blocks that contains a transaction, the wallet sends out 2 notifications.

Test script: https://github.com/patricklodder/dogecoin/blob/1.14.7-walletnotify-test/qa/rpc-tests/walletnotify.py
Results:

```console
$ python3 qa/rpc-tests/walletnotify.py 
Initializing test directory /tmp/test1ytxzzlj/926722
Notifications after reversing:
  d2d825f1cc24e7be6a27e44d11922f17de8ceac833c6ade1b62097b4018584e1 0
  d2d825f1cc24e7be6a27e44d11922f17de8ceac833c6ade1b62097b4018584e1 221
  d2d825f1cc24e7be6a27e44d11922f17de8ceac833c6ade1b62097b4018584e1 0
  d2d825f1cc24e7be6a27e44d11922f17de8ceac833c6ade1b62097b4018584e1 0
  
Assertion failed: 
  File "qa/rpc-tests/test_framework/test_framework.py", line 145, in main
    self.run_test()
  File "qa/rpc-tests/walletnotify.py", line 95, in run_test
    assert len(notifs) == self.current_line + 2
Stopping nodes
Not cleaning up dir /tmp/test1ytxzzlj/926722
Failed
```

I'm okay with not fixing it in this PR (exceeds scope of this change) but we will need to figure out why it does that, and either fix or document the behavior.

PR #3290, opened 2023-07-02
Limit debugfile on SIGHUP

chromatic:
When reading the code and intent around `-shrinkdebugfile`, this omission seemed odd.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -322,6 +322,9 @@ int LogPrintStr(const std::string &str)
             // reopen the log file, if requested
             if (fReopenDebugLog) {
                 fReopenDebugLog = false;
+                if (GetBoolArg("-shrinkdebugfile", !fDebug)) {
+                    ShrinkDebugFile();
+                }
                 boost::filesystem::path pathDebug = GetDataDir() / "debug.log";
                 if (freopen(pathDebug.string().c_str(),"a",fileout) != NULL)
                     setbuf(fileout, NULL); // unbuffered


edtubbs:
> When reading the code and intent around `-shrinkdebugfile`, this omission seemed odd.

Thinking this could work if it doesn't necessarily have to happen at startup https://github.com/dogecoin/dogecoin/blob/1.14.7-dev/src/init.cpp#L470. 

The file could shrink during any of these macros https://github.com/dogecoin/dogecoin/blob/1.14.7-dev/src/util.h#L77C29-L92 but it's only triggered by the SIGHUP signal. So, the user would need to send it via the terminal (e.g., using 'kill') while debugging, right?

chromatic:
We could do it that way, though I'm a little nervous about that. If I have access to your machine but not necessarily your configuration file, logs, wallet, etc then allowing me to send signals to your process to truncate your logs still gives me a nefarious mechanism to manipulate your Core experience.

"But Unky c, why would you give anyone shell access to a Core node even if you've locked down permissions?" That's a good question! That's also a risk. But it's not a strong reason to add a new risk, in my view.

edtubbs:
To make it work a user would need to manually send the SIGHUP signal to the process as core doesn’t do this automatically.

chromatic:
Oh, I see what you mean. Yes, that's correct. Do you think this needs further documentation, or is there something else you're looking for in this PR?

edtubbs:
Consider updating the `-shrinkdebugfile` help message [here](https://github.com/dogecoin/dogecoin/blob/1.14.7-dev/src/init.cpp#L470) to reflect that shrinking occurs via the SIGHUP signal instead of client startup.

chromatic:
@edtubbs I've reviewed the code paths and updated the help text to indicate that shrinking happens on both startup and `SIGHUP`.

edtubbs:
Ok, cool. I'm building this branch on Ubuntu 22.04 with x86_64. I'll test the switch and let you know how it goes.

edtubbs:
Ok here's what I found.. I added some logging to the `ShrinkDebugFile()` and `HandleSIGHUP(int)` functions:

```sh
diff --git a/src/init.cpp b/src/init.cpp
index 75d6cf44a..94efe886c 100644
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -282,6 +282,7 @@ void HandleSIGTERM(int)
 
 void HandleSIGHUP(int)
 {
+    LogPrintf("HandleSIGHUP(int) info, handler called.\n");
     fReopenDebugLog = true;
 }
 
diff --git a/src/util.cpp b/src/util.cpp
index c189b32f3..a1051c710 100644
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -792,7 +792,11 @@ void ShrinkDebugFile()
         }
     }
     else if (file != NULL)
+    {
         fclose(file);
+        LogPrintf("ShrinkDebugFile() info, debug.log within history size.\n");
+    }
+
 }
```
#
Here's the output at startup, confirming that `ShrinkDebugFile` is called successfully during startup:

```sh
ed@ed-VirtualBox:/doge/dogecoin$ ./depends/x86_64-linux-gnu/bin/dogecoin-qt -datadir=./data -debug -debuglogfile=./data/debug.log -shrinkdebugfile -printpriority -printtoconsole

2024-07-11 19:32:31 
...
2024-07-11 19:32:31 Dogecoin version v1.15.0.0-c1d8d616b-dirty
2024-07-11 19:32:31 InitParameterInteraction: parameter interaction: -whitelistforcerelay=1 -> setting -whitelistrelay=1
2024-07-11 19:32:31 Assuming ancestors of block e7d4577405223918491477db725a393bcfc349d8ee63b0a4fde23cbfbfd81dea have valid signatures.
2024-07-11 19:32:31 Prune configured to target 2861MiB on disk for block and undo files.
2024-07-11 19:32:31 Using RdRand as an additional entropy source
2024-07-11 19:32:31 ShrinkDebugFile() info, debug.log within history size.
```
#
On another run, when sending SIGHUP from another console, it logs the signal but `ShrinkDebugFile()` isn't called again:

```sh
ed@ed-VirtualBox:/doge/dogecoin$ ./depends/x86_64-linux-gnu/bin/dogecoin-qt -datadir=./data -debug -debuglogfile=./data/debug.log -shrinkdebugfile -printpriority -printtoconsole

2024-07-11 19:36:02 
...
2024-07-11 19:36:02 Dogecoin version v1.15.0.0-c1d8d616b-dirty
2024-07-11 19:36:02 InitParameterInteraction: parameter interaction: -whitelistforcerelay=1 -> setting -whitelistrelay=1
2024-07-11 19:36:02 Assuming ancestors of block e7d4577405223918491477db725a393bcfc349d8ee63b0a4fde23cbfbfd81dea have valid signatures.
2024-07-11 19:36:02 Prune configured to target 2861MiB on disk for block and undo files.
2024-07-11 19:36:02 Using RdRand as an additional entropy source
2024-07-11 19:36:02 ShrinkDebugFile() info, debug.log within history size.
...
2024-07-11 19:36:03 Opened LevelDB successfully
2024-07-11 19:36:03 Using obfuscation key for /doge/dogecoin/./data/chainstate: a577229fd8dacfd9
2024-07-11 19:36:08 HandleSIGHUP(int) info, handler called.
```
#
Sending the SIGHUP signal from another console:

```sh
ed@ed-VirtualBox:~$ ps -a
    PID TTY          TIME CMD
  10538 tty2     00:00:00 gnome-session-b
  12470 pts/0    00:00:01 dogecoin-qt
  12478 pts/1    00:00:00 ps
ed@ed-VirtualBox:~$ kill -HUP 12470
```
#
So it makes sense to add `ShrinkDebugFile()` to the `HandleSIGHUP(int)` function. After adding that:

```sh
 void HandleSIGHUP(int)
 {
+    LogPrintf("HandleSIGHUP(int) info, handler called.\n");
+    ShrinkDebugFile();
     fReopenDebugLog = true;
 }
```

Then `ShrinkDebugFile` gets called upon sending the SIGHUP signal as expected:

```sh
ed@ed-VirtualBox:/doge/dogecoin$ ./depends/x86_64-linux-gnu/bin/dogecoin-qt -datadir=./data -debug -debuglogfile=./data/debug.log -shrinkdebugfile -printpriority -printtoconsole

2024-07-11 19:46:24 
...
2024-07-11 19:46:24 Dogecoin version v1.15.0.0-c1d8d616b-dirty
2024-07-11 19:46:24 InitParameterInteraction: parameter interaction: -whitelistforcerelay=1 -> setting -whitelistrelay=1
2024-07-11 19:46:25 Assuming ancestors of block e7d4577405223918491477db725a393bcfc349d8ee63b0a4fde23cbfbfd81dea have valid signatures.
2024-07-11 19:46:25 Prune configured to target 2861MiB on disk for block and undo files.
2024-07-11 19:46:25 Using RdRand as an additional entropy source
2024-07-11 19:46:25 ShrinkDebugFile() info, debug.log within history size.
...
2024-07-11 19:46:25 Opened LevelDB successfully
2024-07-11 19:46:25 Using obfuscation key for /doge/dogecoin/./data/chainstate: a577229fd8dacfd9
2024-07-11 19:46:30 HandleSIGHUP(int) info, handler called.
2024-07-11 19:46:30 ShrinkDebugFile() info, debug.log within history size.
```
#
So it seems like `ShrinkDebugFile` needs to be added to `HandleSIGHUP(int)` so it gets called when receiving the signal as you expect.

chromatic:
I see what you're saying, but I'd like to do this a little differently. If we're okay with the shrinking not happening immediately on receiving the signal, I'd like to avoid doing any work _within_ the signal handler (trying to be cognizant of threading here) and shrink the file on the next print to the debug log.

That's this code:

```
            // reopen the log file, if requested
            if (fReopenDebugLog) {
                fReopenDebugLog = false;

                if (GetBoolArg("-shrinkdebugfile", !fDebug)) {
                    ShrinkDebugFile();
                }
                fs::path pathDebug = GetDataDir() / "debug.log";
                if (fsbridge::freopen(pathDebug,"a",fileout) != NULL)
                    setbuf(fileout, NULL); // unbuffered
            }
```

https://github.com/dogecoin/dogecoin/blob/c1d8d616b59b48861213d994dee8a814d12d11f8/src/util.cpp#L324-L334

Does that make sense? So in your testing, you'd have to wait to see another debug line printed (maybe you can force one through an RPC command or wait for a new block or transaction), and then you'll see the file shrink.

edtubbs:
Sure, we can shrink the file during the next debug log print with the mutex there for thread safety. Any reason printing to the console and the debug log have to be mutually exclusive checks?

chromatic:
I think they're exclusive because that's how the `-printtoconsole` CLI option is defined, so changing that might have further implications than they want here. I'm open to code suggestions otherwise, if you see good options.

edtubbs:
Agreed, probably best not to change it since it's documented that way. Here's an updated test.. I've added logging with `FileWriteStr` after `ShrinkDebugFile()` is called:

```diff
diff --git a/src/util.cpp b/src/util.cpp
index c189b32f3..ae1502c58 100644
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -327,6 +327,7 @@ int LogPrintStr(const std::string &str)
                 if (GetBoolArg("-shrinkdebugfile", !fDebug)) {
                     ShrinkDebugFile();
+                    FileWriteStr(LogTimestampStr("Called ShrinkDebugFile()\n"), fileout);
                 }
                 fs::path pathDebug = GetDataDir() / "debug.log";
                 if (fsbridge::freopen(pathDebug,"a",fileout) != NULL)
```

```sh
./depends/x86_64-linux-gnu/bin/dogecoin-qt -datadir=./data -debuglogfile=./data/debug.log -shrinkdebugfile -printpriority
```

```sh
ed@ed-VirtualBox:~$ ps -a
    PID TTY          TIME CMD
   1743 tty2     00:00:00 gnome-session-b
  21419 pts/0    00:00:02 dogecoin-qt
  21428 pts/1    00:00:00 ps
ed@ed-VirtualBox:~$ kill -HUP 21419
```

```sh
2024-07-15 21:10:17 init message: Rewinding blocks...
2024-07-15 21:10:17 Called ShrinkDebugFile()
2024-07-15 21:10:17 init message: Verifying blocks...
2024-07-15 21:10:17 Verifying last 6 blocks at level 3
```

PR #3291, opened 2023-07-04
depends: Update Qt to 5.15.5

alamshafil:
This PR updates the Qt dependency to 5.15.5 and contains the necessary changes for it. 
Qt 5.15.5 was chosen because it is a stable LTS release and is the same version used in Bitcoin.

The way I made this PR was cherry-picking each Bitcoin QT update starting from the version Dogecoin uses. Once I got to the final update, I separated the resulting diff into separate commits.

I have left both the ceb93fa45809d56da176fd96ec4f185f05fbc294 and 6ad6276cfdecbefd23040e8d5f1a673f23e4cc1b commits under `"[WIP]"` because the Qt commit contains many changes that was merged into one. I also am not sure if the 6ad6276cfdecbefd23040e8d5f1a673f23e4cc1b commit was done properly since Bitcoin had completely reworked the file.

I am marking this PR as a draft because it contains many dependency updates such as OpenSSL that can be done seperatly and should be tested. Regrading OpenSSL, Qt 5.15.5 requires a min version of 1.1.1 and I picked 1.1.1s which was the latest at the time.

dogecoin-qt-5-15-5-test
dogecoin-qt-update-qa-test
dogecoin-qt-update


alamshafil:
Looks like I got most of the CI to pass with a couple of failures.

[x86_64-linux-nowallet](https://github.com/dogecoin/dogecoin/actions/runs/5460219086/jobs/9936986593?pr=3291#logs) Looks like Qt 5.15.5 breaks glibc compatibility (I found some features to disable in configure that could fix)

[x86_64-macos](https://github.com/dogecoin/dogecoin/actions/runs/5460219086/jobs/9936986671?pr=3291#logs)
Qt 5.15.5 includes `<os/log.h>` which is only found on >=10.12. Looking further into it, it is trival to get Qt 5.15.5 to work on =10.11, since it requires syscalls only found on >=10.13.

[x86_64-win-experimental](https://github.com/dogecoin/dogecoin/actions/runs/5460219086/jobs/9936986813?pr=3291#logs)
Is this failing because of `bitcoin_qt.m4`? Need to properly fix that file I believe.

alamshafil:
@patricklodder 

Since Qt 5.15.5 will not support macOS 10.11, should we try to get Qt 5.11.x working (supports 10.11) on 1.14.x or just move this to 1.21?

alamshafil:
Since this will break backward compatibility, I will close this PR.

PR #3306, opened 2023-08-03
qa: Add getblock test

alamshafil:
This PR adds a `getblock` RPC test for #3304.

The RPC test works by generating a JSON file with raw block data and output of the `getblock` RPC for each verbosity level. After reading the generated JSON file, which is provided in this PR, the test will startup a node to import all the raw blocks and run the `getblock` RPC to check against the JSON file. Based on `getblockstats` RPC test.


IndigoNakamoto:
👀

eztub:
https://github.com/dogecoin/dogecoin/pull/3306

PR #3307, opened 2023-08-03
[blocked] rpc: Serialize in getblock without cs_main

alamshafil:
For #3304. This PR removes the lock on `cs_main` and only locks it when getting block info. The lock is not present when seriazling block info since data is already fetched in the locked block scope.

Note: #3297 contains a function called `GetBlockChecked()` which can be used here to simplifying checking if a block has been pruned or not found on disk.

I have run this PR with the QA test and JSON data in #3306 and it passed without errors.

**This PR is blocked by #3306 (QA test)**


alamshafil:
Rebased against #3306

PR #3327, opened 2023-08-30
Change Bitcoin to Dogecoin in QA output strings

chromatic:
This is a tiny nit, possibly not worth it, but in reviewing QA debugging output or searching for it online, updating the coin name could remove a small amount of friction and confusion.


chromatic:
I've updated those places as well and scanned the `qa/` directory for any other likely candidates. All that's left is uses of `BITCOIN` and `bitcoind` in the [RPC test runner](https://github.com/dogecoin/dogecoin/blob/1.14.7-dev/qa/pull-tester/rpc-tests.py#L47-L90).

Renaming the environment variable feels more invasive than I'd like; what do you think?

PR #3328, opened 2023-08-31
Add extra configuration file to persist RPC-set features

chromatic:
This begins to address a feature request in #3206. I'm creating it as a draft for now, because we should have make some design decisions before going further.

Current flaws:

- (missing implementation) there's no lockunspent configuration option anywhere
- (code style) the implementation is a mix of C and Boost, so it could be polished further
- (design) this may be the wrong approach
- (potential design question) there's no user notification that we're writing this file apart from its existence
- (potential design question, depends on approach) nothing uses this file

However, in the spirit of debating working code (and in the interest of not replicating uninspectable binary .dat files for important options like these), it's worth at least a draft.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -267,6 +267,7 @@ void Shutdown()
     delete pwalletMain;
     pwalletMain = NULL;
 #endif
+    WriteExtraConfigFile();
     globalVerifyHandle.reset();
     ECC_Stop();
     LogPrintf("%s: done\n", __func__);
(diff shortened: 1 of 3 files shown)


chromatic:
Sorry, pushed the first version of this one commit too early. This at least compiles locally.

chromatic:
I can reproduce this compilation error now locally, so I'll continue working on the code, but I think the design questions remain.

RobRel:
hNbf-oL4M-419T-ksji

PR #3330, opened 2023-09-14
Update GUI to Pepecoin, Pepecoin Addresses start with P

david2278:
Instances of "Dogecoin" strings updated to Pepecoin
Update the GUI for Pepecoin
All Pepecoin addresses now start with P instead of D


david2278:
opened by mistake

PR #3333, opened 2023-09-18
Use C++11 range-based for loops instead of BOOST_FOREACH loop

AjaxPop:
We should prefer C++11 range-based for loops over BOOST_FOREACH loops because they are more efficient and easier to maintain.


chromatic:
I like this change. There are a lot of files where we could do the same; are you interested in updating several others?

AjaxPop:
> I like this change. There are a lot of files where we could do the same; are you interested in updating several others?

Done

AjaxPop:
I am able to get these errors locally. 
I will keep on working on it

AjaxPop:
It looks like my branch that I was working on did not update to the fixate v1.14.6 #3048 version. Also, it looks like some code in the nMockTime update was removed.

Should I just close this pull request out and try again with a smaller number of files?
@patricklodder

AjaxPop:
Given the large number of files that have been changed and the current issue with setmocktime, I am going to close this.

PR #3336, opened 2023-09-22
Remove BOOST_FOREACH

AjaxPop:
You can use the C++11 range-based for loop instead of the Boost BOOST_FOREACH macro for faster iteration.

I'm redoing #3333 code base.

The issue with attempt #3333 was that some files did not update from the git, and as such, it was removing things that someone else added to the codebase.


AjaxPop:
I'm redoing #3333. I'm going go slowly to see which file changed breaks the code base.

PR #3340, opened 2023-10-01
Update build-windows.md, remove linux lib deps installation line.

justinvforvendetta:
Removes misinformation, as this is already on the dogecoin/docs/build-unix.md page. Installing these linux libraries from aptitude can, and will interfere with the libraries that are cross-compiled from /dogecoin/depends/packages m4's using gcc/mingw.

### DIFF
--- a/doc/build-windows.md
+++ b/doc/build-windows.md
@@ -54,10 +54,6 @@ First, install the general dependencies:
     sudo apt upgrade
     sudo apt-get install build-essential libtool autotools-dev automake pkg-config bsdmainutils curl git
     
-If you want to build with the wallet and Qt GUI you also want to install the following (this example is under Ubuntu):
-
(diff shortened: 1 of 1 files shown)


justinvforvendetta:
its a good idea to use aptitude to install the required linux dependencies (but newer versions) just before cross compiling those same deps with the mingw toolchain?

justinvforvendetta:
quite simply, it is not a step in cross compiling the codebase for windows on WSL/Linux, and yes I was referring to Qt having been problematic with things of that nature in the past, and we've all seen Qt bugs go away and then resurface in future versions. Also, experienced users don't need these instructions, and know how to specify paths anyway, but inexperienced users do not, so I made this PR for the greater good. cheers! 

ps: nice to see a very active maintainer here, and I appreciate your vigilance.

justinvforvendetta:
@patricklodder i can help bump some deps versions and linux cross compile updates for building with mingw under ubuntu.. i know much of codebase needs updating, but i can do lil pieces at a time here n there if ya need some help

PR #3384, opened 2023-11-25
doc: added security update guidance

edtubbs:
Split security update guidance into separate PR per https://github.com/dogecoin/dogecoin/pull/3351#issuecomment-1783858659.


patricklodder (approved):
ACK - thank you!

PR #3412, opened 2024-02-11
Disable BIP-70 payment server by default

chromatic:
This commit adds a configuration option, `enable-bip70`, to enable this feature. Only enable this if you know what you're doing.

Please note that BIP-21 payment links continue to work.


patricklodder (commented):
ACK - tested on arm64 MacOS Ventura, x86_64 Ubuntu Jammy and x86_64 Ubuntu Trusty.

Needs final approval from @michilumin because I've contributed to this code

michilumin (commented):
code looks good, overrides look correct and moved code looks appropriately switched.  testing build.

michilumin (approved):
Built and tested to make sure that at least dogecoind -enable-bip70 does indeed execute the load root CA code in PaymentServer,; and does not if not present; so disable/enable logic (with disable default) does seem to be propogating through.

PR #3501, opened 2024-03-29
Update fees.cpp [try-catch]

AjaxPop:
Use a try-catch block for error handling in the TxConfirmStats::Read function.

### DIFF
--- a/src/policy/fees.cpp
+++ b/src/policy/fees.cpp
@@ -3,6 +3,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#include <exception>
 #include "policy/fees.h"
 #include "policy/policy.h"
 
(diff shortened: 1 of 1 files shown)


patricklodder:
What error didn't get caught?

AjaxPop:
Although the check on fileDecay ensures the integrity of the data by confirming that it is between 0 (no decay) and 1 (full decay), there remains a potential issue with data corruption, modification, or a software bug.

The try-catch block ensures that the data integrity adheres to specific expectations regarding its values and structure. Additionally, it has the added benefit of logging out any potential errors.

georgeartem:
Neat. Seems like it would cut out uncertainties.Sent from my iPhoneOn Mar 28, 2024, at 10:27 PM, Skylar Loomis ***@***.***> wrote:﻿
Although the check on fileDecay ensures the integrity of the data by confirming that it is between 0 (no decay) and 1 (full decay), there remains a potential issue with data corruption, modification, or a software bug.
The try-catch block ensures that the data integrity adheres to specific expectations regarding its values and structure. Additionally, it has the added benefit of logging out any potential errors.

patricklodder:
Okay. An overall try-catch already is implemented in the calling code:

https://github.com/dogecoin/dogecoin/blob/6c0095bed1845680b0eee8be53210022773b014c/src/txmempool.cpp#L926

If you're looking to enrich the logged message (I'd support that), it should be done in that catch block instead.

PR #3506, opened 2024-04-02
fix: mising koinu string (rebrand)

danielw86dev:


patricklodder (approved):
ACK. Tested on x86_64 Ubuntu Jammy

This will also remove the ghost entry for translations and unifies it with the one from source:

```console
$ make -C src translate_base
make: Entering directory '/home/patrick/doge/localbuild/src'
  GEN      qt/bitcoinstrings.cpp
  GEN      translate_base
Updating 'qt/locale/bitcoin_en.ts'...
    Found 1075 source text(s) (0 new and 1075 already existing)
    Removed 1 obsolete entries
make: Leaving directory '/home/patrick/doge/localbuild/src'
$ git diff |grep '^[-+].*source'
-        <source>If the custom fee is set to 1000 koinu and the transaction is only 250 bytes, then &quot;per kilobyte&quot; only pays 250 koinu in fee, while &quot;total at least&quot; pays 1000 koinu. For transactions bigger than a kilobyte both pay by kilobyte.</source>
-        <source>If the custom fee is set to 1000 satoshis and the transaction is only 250 bytes, then &quot;per kilobyte&quot; only pays 250 satoshis in fee, while &quot;total at least&quot; pays 1000 satoshis. For transactions bigger than a kilobyte both pay by kilobyte.</source>
+        <source>If the custom fee is set to 1000 koinu and the transaction is only 250 bytes, then &quot;per kilobyte&quot; only pays 250 koinu in fee, while &quot;total at least&quot; pays 1000 koinu. For transactions bigger than a kilobyte both pay by kilobyte.</source>
```

PR #3526, opened 2024-04-18
qt: Display abort node error to UI

alamshafil:
Addresses #3489 Point 1

This PR displays the `AbortNode()` message to the user through a Qt dialog box instead of only printing to the log. 

In addition, the `AbortNode()` logs with `Node aborted: <message>` instead of `*** <message>` to make it clear that the node had to be aborted due to a reason.

Since `AbortNode()` only takes in a single error message now, only the disk space error would be translated. Not sure how this effects translations for other errors. I am also not sure how changing the error string affects the current translations since it remove "Error: " and adds a colon (:) to the end with a space.

dogecoin-error

### DIFF
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -220,7 +220,7 @@ static bool InitRPCAuthentication()
         LogPrintf("No rpcpassword set - using random cookie authentication\n");
         if (!GenerateAuthCookie(&strRPCUserColonPass)) {
             uiInterface.ThreadSafeMessageBox(
-                _("Error: A fatal internal error occurred, see debug.log for details"), // Same message as AbortNode
+                _("A fatal internal error occurred, see debug.log for details: ") + "Unable to open cookie authentication file for writing", // Same message as AbortNode
                 "", CClientUIInterface::MSG_ERROR);
             return false;
         }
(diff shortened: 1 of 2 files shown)


src/httprpc.cpp
```diff
@@ -220,7 +220,7 @@ static bool InitRPCAuthentication()
         LogPrintf("No rpcpassword set - using random cookie authentication\n");
         if (!GenerateAuthCookie(&strRPCUserColonPass)) {
             uiInterface.ThreadSafeMessageBox(
-                _("Error: A fatal internal error occurred, see debug.log for details"), // Same message as AbortNode
+                _("A fatal internal error occurred, see debug.log for details: ") + "Unable to open cookie authentication file for writing", // Same message as AbortNode
```
chromatic, line 223: Should the cookie message be translated too? I understand why this message is split into parts.

jadehamel (commented):
Concatenating strings like this in a translated message might cause issues with localization.

Only part of the message is within the `_()` function, and the additional string is concatenated outside of it.

This can cause issues with localization because the part outside `_()` won't be translated. It's better to include the whole message within the `_()` function and handle any variable content differently.

To avoid localization issues, consider including the entire message within the `_()` function and using placeholders if needed. Here's a suggestion that follows this practice:

```
if (!GenerateAuthCookie(&strRPCUserColonPass)) {
  uiInterface.ThreadSafeMessageBox(
    _("A fatal internal error occurred, see debug.log for details: Unable to open cookie authentication file for writing"), // Same message as AbortNode
```

PR #3613, opened 2024-07-28
Extract set_mocktime() in QA tests

chromatic:
While we probably want to use this helper method on our test classes more broadly, this refactoring is a little invasive because it also requires moving mininode's `wait_until()` helper function and resource lock into the `util` package.

This may be a code smell which suggests further code reorganization before the right design is obvious.


patricklodder (commented):
Looking at the extraction conceptually, I left some comments inline.

PR #3625, opened 2024-08-13
remove repeated words, fix coherency

justinvforvendetta:
remove a few words that are repeated unnecessarily, fix grammar

hope all is well, this week i will be looking at dependencies, and making sure everything is backwards compatible. 

i remember you had a few specific requirements to such, where can i find them patrick?


patricklodder (changes requested):
Please revert changes to `src/secp256k1`, as this is a subtree. You can submit these changes to [bitcoin-core/secp256k1](https://github.com/bitcoin-core/secp256k1) and we'll pick them up with the next update when it's included there.

PR #3664, opened 2024-10-11
doc: fix typos

omahs:
such wow


patricklodder (approved):
utACK - minor overlap with #3525 but that needs some rework, so this is fine.

chromatic:
ACK from me too; all tests pass on x86-64 Linux.

PR #3674, opened 2024-10-31
Add multi-block support and 8-way assembly

edtubbs:
Adds multi-block SHA256 support with ARMv8, AVX2 and SSE optimizations. Integrates Intel’s AVX2 8-way assembly for SHA1, SHA256 and SHA512, with new CI targets and an 8-way config option. SHA1 and SHA512 support to follow. Merge after #3662.


Monireh6569 (commented):
Tnx

P300pl:
Good

P300pl (approved):

PR #3722, opened 2024-11-30
rpc: Add estimatefeereal command

ftab:
Almost like estimatefee but actually works

image

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -435,6 +435,80 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
     return wtx.GetHash().GetHex();
 }
 
+static void PretendMoney(const CTxDestination &address, CAmount nValue, CWalletTx& wtxNew, CAmount& nFeeRequired)
+{
+    CAmount curBalance = pwalletMain->GetBalance();
+
+    // Check amount
+    if (nValue <= 0)
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid amount");
+
+    if (nValue > curBalance)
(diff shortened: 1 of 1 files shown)


ftab:
Some kind of accurate fee determination is desirable for 2nd layer applications using the RPC API. For example I had implemented a Dogecoin wallet in a game running in conjunction with the core wallet and it was a problem to not know in advance how much of a fee it was going to take to send a transaction, so I always had to set `subtractfeefromamount` to true in `sendtoaddress` and this always meant some non-pre-determined amount came out of the requested withdrawal amount and the user could never withdraw a specific amount, say 69.0 doge, into their wallet--it was always 69.0 - fee. With a command like this, I could know the fee in advance of sending the transaction so I could charge the fee to their in-game account separately from the amount being withdrawn.

If non-deterministic transactions are implemented, then I suspect all wallet transactions would have to be that way: subtract fee from amount at the time and you just never know what it will be until the final transaction. Is wallet identification that big of a deal? I never thought so myself but I don't know all the implications.

I don't mind what it's named, `estimatetotalfee` or `determinetotalfee` or anything like that sounds fine to me. I just know that `estimate*fee` as it stands right now are all broken (return -1 as far as I've found) and I did not want to replace one of those commands with one that is incompatible with upstream.

ftab:
> It is, because it leaks information that can be used in targeting for attacks and phishing. Maybe won't work on you or I, but it makes already vulnerable people more vulnerable, so it would be good to be rid of.

Gotcha. Sounds good.

How soon is that coming? Would this best wait for that or can it be useful sooner?

Technical-wise, I'd love a cleaner way to do it than bringing a transaction that close to completion and just not committing, but didn't see an easier way

What should I call the command?

PR #3732, opened 2024-12-02
fix compatibility with boost 1.83

MarsDoge:
[Bugfix] Fix incorrect disconnect call for static functions

Problem:
When using `uiInterface.NotifyBlockTip.disconnect(BlockNotifyGenesisWait)` to disconnect a static function, the compilation fails because `boost::signals2::disconnect` requires the exact type signature match of the slot. Using the function name directly doesn't resolve to the correct function pointer for static functions.

Solution:
The fix explicitly uses `&BlockNotifyGenesisWait` to pass the function pointer. This resolves the compilation error.

Impact:
Only affects the disconnect mechanism for static functions in `uiInterface.NotifyBlockTip`.


MarsDoge:
@patricklodder Can you review this patch for me?

MarsDoge:
Looking forward to your review and test results.

MarsDoge:
This test looks good.

PR #3782, opened 2025-02-07
fix: typos in documentation files

maximevtush:


chromatic:
This is a pretty small change; is checking as if it were Norwegian sufficient?

src/qt/locale/bitcoin_nb.ts
```diff
@@ -3935,7 +3935,7 @@ Use this functionality with extreme caution.</source>
     </message>
     <message>
         <source>Tor control port password (default: empty)</source>
-        <translation>Passord for Tor-kontrollport (standardverdi: tom)</translation>
+        <translation>Password for Tor-kontrollport (standardverdi: tom)</translation>
```
thisiskeanyvy, line 3938: I understand you're proposing to change "passord" to "password" in the Norwegian translation, but I must advise against this modification. "Passord" is the correct Norwegian term for "password", while "password" is the English word. Norwegian, like many languages, adapts technical terms into its own vocabulary rather than directly using English words. Using "passord" is not only linguistically accurate but also consistent with Norwegian software localization standards. While many Norwegians might understand "password" due to English influence, it's not considered correct in formal Norwegian or official contexts. Maintaining "passord" ensures the translation remains idiomatic and adheres to Norwegian language conventions for technical terminology.

src/qt/locale/bitcoin_nb.ts
```diff
@@ -3983,7 +3983,7 @@ Use this functionality with extreme caution.</source>
     </message>
     <message>
         <source>Password for JSON-RPC connections</source>
-        <translation>Passord for JSON-RPC forbindelser</translation>
+        <translation>Password for JSON-RPC forbindelser</translation>
```
thisiskeanyvy, line 3986: Same thing here.

thisiskeanyvy (changes requested):
I don't think this translation is necessary, especially as it is wrong.

You can check in the official Bokmål dictionary : [https://naob.no/ordbok/passord](https://naob.no/ordbok/passord)

PR #3795, opened 2025-03-01
Add configure option to warn on NULL/0 usage

chromatic:
This will help us migrate from checking for NULL to checking for NULLPTR. Enable this at configuration time with:

    ./configure --enable-null-warnings
    
 With this option enabled, I see compilation warnings like:
 
 ```
 ./qt/peertablemodel.h: At global scope:
./qt/peertablemodel.h:49:51: warning: zero as null pointer constant [-Wzero-as-null-pointer-constant]
   49 |     explicit PeerTableModel(ClientModel *parent = 0);
```

All tests pass for me with this flag enabled (as expected).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -69,6 +69,12 @@ AC_ARG_ENABLE([c++17],
   [use_cxx17=$enableval],
   [use_cxx17=no])
 
+AC_ARG_ENABLE([null-warnings],
+  [AS_HELP_STRING([--enable-null-warnings],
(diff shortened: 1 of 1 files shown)


chromatic:
Fair point; the only real argument here is I tend to forget what I've added as an environment variable when running configure, so this approach lets me pick it up again but given that it's a developer-only configuration-time option, your laziness point is persuasive.

chromatic:
There's no label "do not merge (until Patrick and I can discuss)", so I used the best we had.

PR #3796, opened 2025-03-09
Replace boost::function with std::function

chromatic:
Given support for C++-11, we can start migrating some dependencies to the standard library.

### DIFF
--- a/src/bench/bench.h
+++ b/src/bench/bench.h
@@ -7,7 +7,7 @@
 
 #include <cstdint>                               // for uint64_t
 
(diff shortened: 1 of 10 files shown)


chromatic:
Are you using `iwyu` or something else?

chromatic:
Thanks for digging into that! I'm happy to make any other similar changes.

I'm nervous about pulling on the iwyu threads too much, lest we destroy our sweaters. That'd make a good followup though, especially as we can adopt more and more C++ features and retire their Boost predecessors.

PR #3816, opened 2025-04-02
docs: added a badge to the CodeQL workflow

bravesasha:


bravesasha:
I understand, thanks for the feedback.

PR #3827, opened 2025-04-19
Quiet Compiler Warnings about Uninitialized Data

chromatic:


chromatic:
I'm willing to try that. Way back at the start I did try explicitly initializing the union but that caused further problems so I ended up trying the approach, but the trivially copyable assertion you found is intriguing.

chromatic:
I managed a clean run with something like this instead:

```
diff --git a/src/prevector.h b/src/prevector.h
index cba2e3005..7f296d2a3 100644
--- a/src/prevector.h
+++ b/src/prevector.h
@@ -219,7 +219,7 @@ public:
         }
     }
 
-    prevector() : _size(0) {}
+    prevector() : _size(0), _union{} {}
 
     explicit prevector(size_type n) : _size(0) {
         resize(n);
@@ -254,7 +254,7 @@ public:
         }
     }
 
-    prevector(prevector<N, T, Size, Diff>&& other) : _size(0) {
+    prevector(prevector<N, T, Size, Diff>&& other) : _size(0), _union{} {
         swap(other);
     }
 
@@ -274,7 +274,11 @@ public:
     }
 
     prevector& operator=(prevector<N, T, Size, Diff>&& other) {
-        swap(other);
+        if (this != &other) {
+            _union = {};
+            _size = 0;
+            swap(other);
+        }
         return *this;
     }
```

... which I like a *lot* better than `memset` tricks, especially for readability.

patricklodder:
> ... which I like a lot better than memset tricks, especially for readability.

Definitely improves readability. I have time to spend on a proposal, could you open a PR (or update this one?)

PR #3836, opened 2025-05-02
Update OP_RETURN limit to 420

UsaRandom:
From discussion: https://github.com/dogecoin/dogecoin/discussions/3829

This changes the standard transaction definition to include OP_RETURN utxos up to 420 bytes.

```cpp
static const unsigned int MAX_OP_RETURN_RELAY = 424; //!< bytes (+1 for OP_RETURN, +1 for the pushdata opcode, +2 for pushdata length)```


FierceSkit:
NACK

After reviewing the points, that had been made in #3829 considering the "pro" and cons:
I cannot see a necessity or benefit for anyone but -at "best"- a very small group of people.
As of now, all I see is the potential for unnecessary bloat, which brought me to the same conclusion as:

**"DOGECOIN IS NOT A STORAGE NETWORK! IF YOU WANT TO STORE FILES, TEXT MESSAGES OR OTHER BLOAT, ASK DEEPSEEK TO CODE YOU A GOOD NETWORK SOFTWARE, FFS."**

Yes, you can quote me on that(aswell).

John-Inubook:
I agree with @FierceSkit, I am still not convinced for the real utility of this and Dogecoin is not a storage network.

John-Inubook:
Wouldn't it affect scalability and make running a node pricier and slower?

PR #3840, opened 2025-05-04
rpc: unify auxpow caching / simplify implementation

### LINKED ISSUE #3828: [bug] createauxblock and getauxblock simultaneously lead to random behaviour

patricklodder:
Unifies the implementation of auxpow block template methods on the RPC, and with that, fixes #3828.

1. Introduces a single cache for templated blocks `CAuxBlockCache`, that is indexed and searchable by both coinbase script and block hash. This replaces multiple separate caches implemented in `std::map` structures.
2. Integrates the new cache into `AuxMiningCreateBlock` and `AuxMiningSubmitBlock` rpc helper functions
3. Modifies `AuxMiningSubmitBlock` to return the BIP-22 response
4. Merges `getauxblockbip22` into `getauxblock`, to decrease code complexity.
5. Improves error handling on submitting auxpow blocks on all rpc methods, failing safely and early, with improved error messages.

Note: although we return the BIP-22 response in `AuxMiningSubmitBlock`, there is currently no method or configuration that returns this over rpc. All methods return a boolean response and this remains unchanged, for backward compatibility.

### DIFF
--- a/src/rpc/auxcache.cpp
+++ b/src/rpc/auxcache.cpp
@@ -1,0 +1,135 @@
+// Copyright (c) 2025 The Dogecoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "rpc/auxcache.h"
+
+#include "script/standard.h"  // for CScriptID
+#include "primitives/block.h" // for CBlock
+#include "uint256.h"          // for uint256
+#include "utilmemory.h"       // for MakeUnique
+#include "utiltime.h"         // for GetTimeMicros
+
+#include <boost/multi_index_container.hpp>
+#include <boost/multi_index/ordered_index.hpp>
+
+#include <cstdint>            // for uint64_t
+#include <memory>             // for std::unique_ptr, std::shared_ptr
+#include <mutex>              // for std::mutex
+#include <tuple>              // for std::tuple
+
+namespace {
+
+//! Type that is stored in the cache
+struct AuxCacheItem {
+  int64_t cache_time;
+  CScriptID scriptId;
+  uint256 hash;
+  std::shared_ptr<CBlock> pblock;
+};
+
+// The ByScriptId index sorts by CScriptID and item age
+struct ByScriptId {};
+using ByScriptIdView = std::tuple<const CScriptID&, const int64_t>;
+struct ByScriptIdExtractor
+{
+    using result_type = ByScriptIdView;
+    result_type operator()(const AuxCacheItem& item) const
+    {
+        // calculate age
+        const int64_t age = GetMockableTimeMicros() - item.cache_time;
+        return ByScriptIdView{item.scriptId, age};
+    }
+};
+
+// The ByBlockHash index sorts by blockhash (uint256)
+struct ByBlockHash {};
+struct ByBlockHashExtractor
+{
+    using result_type = uint256;
+    result_type operator()(const AuxCacheItem& item) const
+    {
+        return item.hash;
+    }
+};
+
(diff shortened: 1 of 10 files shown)


src/rpc/auxcache.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+};
+
+/** Data type for the main data structure (AuxCacheItem objects with ByScriptID/ByBlockHash indexes). */
+using AuxCacheIndex = boost::multi_index_container<
+    AuxCacheItem,
+    boost::multi_index::indexed_by<
+        boost::multi_index::ordered_non_unique<boost::multi_index::tag<ByScriptId>, ByScriptIdExtractor>,
+        boost::multi_index::ordered_unique<boost::multi_index::tag<ByBlockHash>, ByBlockHashExtractor>
+    >
```
chromatic, line 63: It would be nice if C++ someday offered an STL replacement for this, but Boost it is for now!

src/rpc/auxpow.cpp
```diff
@@ -150,165 +139,22 @@ static UniValue AuxMiningCreateBlock(const CScript& scriptPubKey)
     return result;
 }
 
-static bool AuxMiningSubmitBlock(const std::string& hashHex, const std::string& auxpowHex)
+static UniValue AuxMiningSubmitBlock(const std::string& hashHex, const std::string& auxpowHex)
```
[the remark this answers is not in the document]
chromatic: I'm 100% on board with that in a separate commit. Great idea!

src/rpc/auxpow.cpp
```diff
@@ -62,14 +57,12 @@ void AuxMiningCheck()
 
 static UniValue AuxMiningCreateBlock(const CScript& scriptPubKey)
 {
-
     AuxMiningCheck();
-    LOCK(cs_auxblockCache);
+    LOCK(cs_auxpowrpc);
 
-    static unsigned nTransactionsUpdatedLast;
+    static unsigned int nTransactionsUpdatedLast;
```
chromatic, line 63: Thank you for cleaning this up along the way.

src/rpc/auxpow.cpp
```diff
@@ -150,19 +139,20 @@ static UniValue AuxMiningCreateBlock(const CScript& scriptPubKey)
...
 
-    const std::map<uint256, CBlock*>::iterator mit = mapNewBlock.find(hash);
-    if (mit == mapNewBlock.end())
+    std::shared_ptr<CBlock> pblock;
+    bool maybe_block = auxBlockCache.Get(hash, pblock);
+    if (!maybe_block) {
         throw JSONRPCError(RPC_INVALID_PARAMETER, "block hash unknown");
-    CBlock& block = *mit->second;
+    }
+    CBlock& block = *pblock;
```
chromatic, line 154: I'm struggling a bit to figure out the shared pointer lifecycle here; we should rely on cache expiration/clearing to clean this up here, right? No need for an explicit `reset()` in this codepath?

chromatic (commented):
All tests pass. Good to go with a squash (and possibly a response about my pblock lifecycle comment).

src/rpc/auxpow.cpp
chromatic: Oh right, copy on assignment here. Sorry, I should have re-read the class operators before asking. This all makes sense. I was reading the code closely to make sure we handled the lifecycle correctly so this codepath jumped out at me.

chromatic (approved):

PR #3842, opened 2025-05-12
Update dogecoin.cpp

AjaxPop:
Instead of using Boost, use C++11 <random> instead.


chromatic:
Looks like we use these Boost libraries also in `src/test/scheduler_tests.cpp`, `src/tests/random_tests.cpp`, and `src/dogecoin-fees.cpp`. Can those files be updated as part of this PR as well?

PR #3844, opened 2025-05-20
Add OP_RETRUN filters to dogecoin.conf

qlpqlp:
I added this approach that I already talked a few years ago that could maybe help, because its completely optional to any shibe runing a Dogecoin node.

on **dogecoin.conf** file you can add filters like so:

```
# To block Doginals
filterdoginals=1

# To Block DRC-20
filterdrc20=1

# To add custom filters
opreturnfilter=your-pattern-here-1
opreturnfilter=your-pattern-here-2
```

and would on that node reject on mempool any transactions with that pattern


0foxes (approved):
Builds and seems to work properly for me. +1 as I think the option would be good to have for shibes.

Pimax1:
Optional OP_RETURN filters could be seen as censorship on Dogecoin’s permissionless network, especially given the strong anti-Doginals sentiment behind this proposal. 
Are you planning to keep submitting PRs to update a list of banned "words" based on personal dislike? That could become unmanageable and fragment the network.

qlpqlp:
> There already is a config option for `OP_RETURN` policy:
> 
> ```
>   -datacarrier
>        Relay and mine data carrier transactions (default: 1)
> 
>   -datacarriersize
>        Maximum size of data in data carrier transactions we relay and mine
>        (default: 83)
> ```

Did not knew about that 😅
But, for my understanding it only limits size, don't filter, correct?

chromatic:
I'm nervous about encoding code for a specific policy choice here, given three things:

- developers should be wary of network fragmentation or influencing any types of transactions
- there's already a mechanism which accomplishes something similar (not the same, but similar)
- if shoving data past `OP_RETURN` is no longer feasible (if this approach works), people who really want to store non-transaction data on the blockchain will find another approach, and I'm reluctant to get into an arms race


With that said, I think the code as written here is inside-out; I'd prefer to perform the flag check first, then examine the transaction. That way the cost of looking for the offending transactions is only on people who use this flag.

Pimax1:
> Side note: did y'all really test this or you're bluffing? I can't help but feel that you're hard forking here, that doginals don't use OP_RETURN nor does it use scriptPubKey (so you're filtering the wrong field and adding the wrong conditions.)

Kudos to @patricklodder !  OP_RETURN filter for Doginals/DRC-20 is a complete miss.
none of these transactions use OP_RETURN, so it doesn't filter a single one of them.

Looking at some @qlpqlp’s X posts makes this PR seem like a deliberate attempt to stir drama and fracture the community rather than improve Dogecoin and a waste of everyone's time  :angry:

PR #3860, opened 2025-06-30
Add Kurdish translation of README

FouadAbdi:


thisiskeanyvy:
The structure and translations seem to be faithful. However, a review by a native speaker wouldn't hurt and perhaps the name of the Ku folder should be changed to ku I think to reflect the current repo structure.

sabasarmay (commented):
great this kurmanci translateion is good

PR #3917, opened 2025-10-20
src/utils/dogecoin_version_check.cpp

Motelek12:
g++ src/utils/dogecoin_version_check.cpp -o doge-version
./doge-version
Dogecoin Core version: 1.14.9


georgeartem:
you wanted to print this to the console? why?

PR #3919, opened 2025-10-21
Add small utility to print Dogecoin Core version for debugging

iruk99090:
Developers and testers often need a simple way to confirm which version of Dogecoin Core is running when debugging local builds or CI runs.
This helper makes it easy to print that info programmatically.

Impact:

No dependencies or functional changes

Compatible with all current build systems

Useful for development and diagnostics


georgeartem:
if it's unplanned for a release it won't get done - what was frustrating about not being able to print to console?

you needed a sanity check to be sure which node version you were running?

it would be nice to be able to have more utils but why this specifically?

PR #3720, opened 2024-11-29
Fix be2enc and be2dec includes

chromatic:
Attempt to address #3673 . This approach works for me on x86-64 Linux.

PR #3930, opened 2025-10-26
compat: fix endian.h compatibility with BSD(-like) systems

patricklodder:
Re-do of #3720, copying the logic from boost to specifically include `sys/endian.h` for FreeBSD and `machine/endian.h` for MacOS, to reduce conflicts. Also see https://github.com/dogecoin/dogecoin/pull/3720#issuecomment-2517335234.

Fixes inclusion between the different endian.h implementations:

- endian.h (GNU, OpenBSD)
- sys/endian.h (FreeBSD)
- machine/endian.h (MacOS)

exactly as done by boost_predef, so that we do not get conflicts where symbols get overridden. Even though a system endian.h may be available of another flavor than included by boost, this does not help us with redefinitions, so we limit ourselves to using the boost method exclusively, until we are free of dependencies to boost_predef.

Also moves the declarations to compat/endian.h instead of trying to solve this in scrypt-specific code.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -607,19 +607,29 @@ if test x$TARGET_OS = xdarwin; then
   AX_CHECK_LINK_FLAG([[-Wl,-dead_strip]], [LDFLAGS="$LDFLAGS -Wl,-dead_strip"])
 fi
 
-AC_CHECK_HEADERS([endian.h sys/endian.h byteswap.h stdio.h stdlib.h unistd.h strings.h sys/types.h sys/stat.h sys/select.h sys/prctl.h])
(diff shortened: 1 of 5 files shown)


chromatic (commented):
Concept ACK.

tojin12341za (approved):
@Owner

patricklodder:
Tested to fix #3975 on MacOS with:

- clang `arm64-apple-darwin25.3.0`
- Xcode `26.0`

so this is broader than FreeBSD now.

tojin12341za (approved):

PR #3932, opened 2025-10-28
Dogecoin RPC example

kilfoigo12:
This PR adds a minimal DogecoinRpcExample.cpp under contrib/ demonstrating how to connect to a local Dogecoin Core node using JSON-RPC.

Purpose:

Quickly verify Dogecoin node RPC connectivity

Lightweight, standalone example for new contributors

Easy to extend to other RPC commands or testnet/regtest

Changes:

Added DogecoinRpcExample.cpp

Uses curl to send JSON-RPC requests

Shows example with getblockcount

### DIFF
--- a/DogecoinRpcExample.cpp
+++ b/DogecoinRpcExample.cpp
@@ -1,0 +1,32 @@
+#include <iostream>
+#include <curl/curl.h>
+#include <string>
+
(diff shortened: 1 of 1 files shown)


kilfoigo12:
1@

PR #3946, opened 2025-11-27
Update ax_boost_filesystem.m4

LeanAira10:


chromatic:
I'm leery of removing copyright and provenance information from this file. What motivated these changes? It's difficult for me to review this request without understanding that more.

patricklodder:
I suspect that it's a duplicate of #3928 in terms of the problem it implies to solve.

Either way, this file is imported from the [GNU autoconf archive](https://www.gnu.org/software/autoconf-archive/The-Macros.html) and we shouldn't be modifying it unless we have a good reason to.

PR #3962, opened 2026-01-14
docs: Update vMerkleBranch comment in auxpow.h

EricGrill:
## Summary

Replaces an outdated TODO comment in `auxpow.h` that questioned whether `vMerkleBranch` was used.

**Before:**
```cpp
// Dogecoin TODO: Is this used? If not remove. If it is, I don't think it's actually set
// anywhere. Check with Namecore
std::vector<uint256> vMerkleBranch;
```

**After:**
```cpp
/** Merkle branch connecting this transaction to the block's merkle root.
 *  Set by InitMerkleBranch() and verified in CAuxPow::check(). */
std::vector<uint256> vMerkleBranch;
```

## Context

The field IS actively used:
- **Set by** `InitMerkleBranch()` at `auxpow.cpp:42`
- **Verified in** `CAuxPow::check()` at `auxpow.cpp:100`  
- **Serialized** in `CMerkleTx::SerializationOp()` at `auxpow.h:83`

The TODO comment was outdated and potentially misleading for future contributors.

## Test plan

- [x] Verify the field is actually used (grep confirmed)
- [x] Build passes (documentation-only change)

---
Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/auxpow.h
+++ b/src/auxpow.h
@@ -36,8 +36,9 @@ private:
 public:
     CTransactionRef tx;
     uint256 hashBlock;
-    // Dogecoin TODO: Is this used? If not remove. If it is, I don't think it's actually set
-    // anywhere. Check with Namecore
+
+    /** Merkle branch connecting this transaction to the block's merkle root.
+     *  Set by InitMerkleBranch() and verified in CAuxPow::check(). */
     std::vector<uint256> vMerkleBranch;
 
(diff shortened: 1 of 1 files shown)


chromatic (approved):
ACK; verified comment and description. All tests pass on x86-64 Linux.

Thank you for adding file names and line numbers in the description; that made it easy to check your work!

PR #3968, opened 2026-01-28
update old xanimo-key

xanimo:
i no longer have access to the previous key so opting for replacement.


patricklodder (approved):
Validated this on a separate channel. I'll merge this in a few hours from now.

PR #3970, opened 2026-02-06
src, config, depends, test: remove openssl

edtubbs:
Remove OpenSSL and perfmon seeding, use OS RNG directly, and add a local `memory_cleanse`.


victorsk2019:
Hi!  Does this mean we may not need openssl dependency when compiling Dogecoin Core with this PR?

edtubbs:
We’re taking the same approach bitcoin did to drop openssl (https://github.com/bitcoin/bitcoin/pull/17265) and windows docs say CryptGenRandom uses a system wide seeding (https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptgenrandom), so dropping perfmon seeding like bitcoin did makes sense too (https://github.com/bitcoin/bitcoin/pull/31124). 

I think we should still layer on the backport in #3522.

xanimo:
> > use OS RNG directly
> 
> 2 questions:
> 1. Is this safe? There's been a lot of issues with insecure randomization in the past, so I'd like to know why this is okay and we don't need a seeding RNG.
> 2. Does this mean we're giving up on #3522? @xanimo?

i'm not giving up on #3522 unless it's rejected.

chromatic:
I agree with @patricklodder we should get something like #3522 in place and see how that works before we're ready to remove OpenSSL altogether. @xanimo how can we help there?

xanimo:
#3522 is the substantive piece. it's closed; of its bases #3466, #3511 and #3521 are in, #3487 and #3520 aren't.

reopening it with what's left after #3521 forward-ported onto master: the PRNG switch and the strengthening loop. a rebase doesn't work, master reached the `__get_cpuid` form of the rdrand check by a different route so replaying the 2024 history asks it to go backwards.

both rewrite random.cpp, but the collision that matters is `CInit`. this PR deletes the openssl locking block from util.cpp and #3522 moves that same block into random.cpp, so git won't flag it and whichever lands second gets rewritten by hand.

two things here sit outside that scope. `RAND_event` in winshutdownmonitor.cpp is bitcoin cc3b5289e byte for byte apart from the util/system.h include, and that commit is already on 1.21-dev, so it can carry its own message rather than land inside the openssl removal.

the other is support/cleanse.cpp. memset plus `atomic_signal_fence` plus the asm barrier isn't upstream's shape: bitcoin has never had the fence in that file, and since c399c80a0 it branches on WIN32 to `SecureZeroMemory` for mingw-w64 release builds, which is how we ship windows.

#3522 first leaves this one as build plumbing plus those two.

xanimo:
correcting what i wrote above. as #3522 stands it doesn't leave this one as build plumbing, it grows it. following 14955 means openssl moves into RNGState rather than out: conf.h included, the locking callbacks and RAND_cleanup as members, RAND_bytes and RAND_screen as inputs to the mixer. that's what upstream did, and #17265 is the pr that undid it.

the overlap to watch is that both branches delete the same 51 lines from util.cpp at the same offset, so git merges them with no conflict, and CRYPTO_set_locking_callback and RAND_cleanup end up live in random.cpp against a tree where depends has dropped libcrypto. RandAddSeedPerfmon is the same shape: #3522 keeps and reworks it, this one deletes it. neither shows up in either diff.

PR #3978, opened 2026-03-21
validation: fix incorrect fee value in mempool rejection message

cuiweixie:
The error message for "mempool min fee not met" reported nFees instead of nModifiedFees, which is the value actually compared against mempoolRejectFee. This made the logged rejection reason misleading when fee deltas were applied, as the reported value did not match the value used in the comparison.


chromatic (approved):
ACK, all tests pass on x86-64 Linux. Thank you for the keen eyes.

patricklodder:
Added this to the backport list

PR #3979, opened 2026-03-21
dogecoin: remove unused bnOld in retarget calculation

cuiweixie:
bnOld was assigned from nBits but never read; drop the dead store.


chromatic (approved):
ACK, all tests pass on x86-64 Linux. Thank you.

patricklodder (approved):
utACK - confirmed that this was an artifact of the rewrite in 1.10, 8da45ed

I've made a note to look into enhancing `src/test/dogecoin_tests.cpp` coverage a little bit more because it's on the light side. Isn't blocking for this PR though.

PR #3985, opened 2026-04-09
Minor typo fixes, comment wording and style consistency cleanups

pmzajaczkowski:
This PR collects three small, non-functional cleanup commits across tools,
tests, comments and source files. Each commit is isolated by kind of change.

**Commit 1 – Fix typos in tools and tests**
- `contrib/devtools/optimize-pngs.py`: fix a typo in a comment.
- `src/warnings.cpp`: rename a misspelled local variable (`warningmessage` →
  corrected spelling). No behavioral change.
- ~~`src/test/transaction_tests.cpp`: remove a duplicate `#include` line.~~ (removed upon request, will be addressed separately)

**Commit 2 – Improve wording in comments and docs**
Rephrase a small number of in-code comments and one developer-notes entry to
make the language more natural and easier to read:
- `doc/developer-notes.md`
- `src/bench/coin_selection.cpp`
- `src/bench/mempool_eviction.cpp`
- `src/miner.cpp`
- `src/wallet/wallet.cpp`

Technical meaning is unchanged in every case.

**Commit 3 – Apply small style consistency cleanups**
Align a handful of spots with the style conventions described in
`doc/developer-notes.md`:
- `src/coins.h`: normalize spacing inside a Doxygen `/** … */` block
  (`@param` / `@return` lines used inconsistent tab/space mixing).
- `src/addrman.cpp`: move opening brace of `RandomInt` to its own line.
- `src/netmessagemaker.h`: expand a single-line constructor definition to the
  multi-line brace style used elsewhere in the file.

PR #3996, opened 2026-05-07
Remove duplicate #include lines in three source files

pmzajaczkowski:
This PR removes duplicate `#include` directives found across three source files. The duplicates are purely mechanical — no header is intentionally included twice, and removing the second occurrence has no behavioral effect.

The change to `src/test/transaction_tests.cpp` was originally part of [#3985](https://github.com/dogecoin/dogecoin/pull/3985) but was moved to a separate PR following reviewer feedback that it fit better on its own.

**Changes:**
- `src/test/transaction_tests.cpp` — remove duplicate `<boost/assign/list_of.hpp>`
- `src/httprpc.cpp` — remove duplicate `"utilstrencodings.h"`
- `src/qt/rpcconsole.cpp` — remove duplicate `"bantablemodel.h"` and `<QThread>`

A duplicate was also found in `src/leveldb/port/port_stdcxx.h`, but that file is part of the vendored LevelDB library and is left unchanged.

### DIFF
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -16,7 +16,6 @@
 #include "ui_interface.h"
 #include "crypto/hmac_sha256.h"
 #include <stdio.h>
-#include "utilstrencodings.h"
 
 #include <boost/algorithm/string.hpp> // boost::trim
 #include <boost/foreach.hpp> //BOOST_FOREACH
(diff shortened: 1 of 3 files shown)


pmzajaczkowski:
The CI failure in the `x86_64-linux-dbg` job is in `fundrawtransaction.py` and appears unrelated to this change. The test fails during node shutdown with a `ConnectionRefusedError` — one of the nodes terminated before the RPC `stop` call could complete. This change only removes duplicate `#include` lines and does not touch transaction handling logic. Could a maintainer re-run CI to confirm, please?

chromatic (approved):
ACK; all tests pass on x86-64 Linux.

PR #3997, opened 2026-05-07
cleanup: remove trailing whitespace across src/

pmzajaczkowski:
This PR strips trailing whitespace from a number of source files across `src/`.

The project has a `.clang-format` configuration and a `clang-format-diff.py`
helper in `contrib/devtools/`, but there is no CI step that enforces whitespace
hygiene automatically. This cleanup removes the existing issues so that any
future linting or editor integration starts from a clean baseline.

No logic, structure, or code meaning is affected.

Vendored subtrees (`src/leveldb/`, `src/secp256k1/`) are intentionally excluded.


chromatic (approved):
ACK; all tests pass on x86-64 Linux.

PR #3998, opened 2026-05-07
ci: add source code linting to GitHub Actions, replace Travis CI docs

pmzajaczkowski:
This PR extends the existing `linter.yml` workflow with two new source code
checks, and updates the documentation to reflect that the project now uses
GitHub Actions instead of Travis CI.

**Changes:**

*Commit 1 — trailing whitespace check*
Adds a CI job that scans all C++ source files under `src/` for trailing
whitespace and fails if any are found. Vendored subtrees (`leveldb`,
`secp256k1`) are excluded. The pattern covers both trailing spaces and tabs.

*Commit 2 — clang-format check*
Adds a CI job that runs `clang-format --dry-run --Werror` against all C++
files changed in the PR, using the project's existing `src/.clang-format`
configuration. The base ref is resolved correctly for both `pull_request`
and `push` events.

The project already ships `src/.clang-format` and recommends using
`clang-format-diff.py` before submitting PRs (see `doc/developer-notes.md`),
but this was never enforced in CI. This commit closes that gap.

*Commits 3–4 — documentation*
Replace references to Travis CI with GitHub Actions across
`doc/developer-notes.md` and `doc/README.md`, add a new
`doc/github-actions-ci.md` describing the current workflow setup, and
remove the now-obsolete `doc/travis-ci.md`.

---

**Note:** PR #3997 (trailing whitespace cleanup across `src/`) should be
merged before this one. Commit 1 adds a check that would fail on the
current codebase — #3997 cleans up the existing violations so this check
can go green from day one.


pmzajaczkowski:
The `whitespace` job in the linter workflow is expected to fail on this PR — it was added as part of this change and the codebase currently contains trailing whitespace in several source files. PR #3997 cleans up all existing violations and should be merged first. Once #3997 lands, this check will pass cleanly from the start. Could we get #3997 reviewed and merged before re-running CI here, please?

.github/workflows/linter.yml
```diff
@@ -30,3 +30,31 @@ jobs:
...
+      with:
+        fetch-depth: 0
+    - name: Install clang-format
+      run: sudo apt update && sudo apt install -y clang-format
+    - name: Check formatting of changed source files
+      env:
+        BASE_SHA: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || 'HEAD~1' }}
+      run: |
+        git diff --diff-filter=d --name-only "$BASE_SHA" -- '*.cpp' '*.h' '*.cc' '*.cxx' | \
+          xargs -r clang-format --dry-run --Werror -style=file
```
patricklodder: Would be better to use the python script from `contrib/devtools` here too, for consistency.

pmzajaczkowski: @patricklodder OK

chromatic:
@pmzajaczkowski I added the "rebase needed" label because merging it now might fail, if we've added more code since merging #3997 . Would you mind rebasing so we can see if we need to do another cleanup round before merging this?

I realize merging this might invalidate other PRs in progress, but we can address that in those PRs after we merge this one.

PR #4012, opened 2026-06-10
p2p: mitigate remote header DoS (side-fork storage + unconnecting-header bypass)

ouicate:
## Summary

Addresses a combined P2P header-processing denial-of-service vulnerability with two root causes:

1. **Unbounded `mapBlockIndex` growth** — valid post-checkpoint side-fork headers could be accepted indefinitely with no global cap or eviction.
2. **`MAX_UNCONNECTING_HEADERS` bypass** — `nUnconnectingHeaders` was reset to zero after *any* accepted connecting header, including unrelated low-work side forks, allowing attackers to avoid escalating misbehavior penalties.

### Changes

- **Side-fork resource limit** (`validation.cpp` / `validation.h`): reject new header-only entries on low-work side forks once `MAX_LOW_WORK_SIDEFORK_HEADERS` (2048) is reached. A header is treated as a useful extension when it extends the active tip, extends the best-header chain during sync, or meets/exceeds best known chain work; competing forks from earlier active-chain blocks are capped.
- **Per-peer side-fork quota** (`net_processing.cpp`): score peers that exceed `MAX_LOW_WORK_SIDEFORK_HEADERS_PER_PEER` (128).
- **Conditional `nUnconnectingHeaders` reset** (`net_processing.cpp`): only reset when accepted headers are on the active chain, compete with tip work, extend the best-header chain, or resolve a previously announced unknown block — not for arbitrary low-work side forks.
- **Regression tests** (`src/test/headers_dos_tests.cpp`): verify the global cap rejects excess competing regtest headers while main-chain extensions remain accepted.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -191,6 +191,8 @@ struct CNodeState {
     const CBlockIndex *pindexBestHeaderSent;
     //! Length of current-streak of unconnecting headers announcements
     int nUnconnectingHeaders;
+    //! Low-work side-fork headers accepted from this peer since connect
+    int nLowWorkSideForkHeadersFromPeer;
     //! Whether we've started headers synchronization with this peer.
     bool fSyncStarted;
     //! When to potentially disconnect peer for stalling headers download
@@ -233,6 +235,7 @@ struct CNodeState {
         pindexLastCommonBlock = NULL;
         pindexBestHeaderSent = NULL;
         nUnconnectingHeaders = 0;
+        nLowWorkSideForkHeadersFromPeer = 0;
         fSyncStarted = false;
         nHeadersSyncTimeout = 0;
         nStallingSince = 0;
@@ -2455,6 +2458,13 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
         }
         }
 
+        uint256 hashPrevUnknown;
+        {
+            LOCK(cs_main);
+            CNodeState *nodestate = State(pfrom->GetId());
+            hashPrevUnknown = nodestate->hashLastUnknownBlock;
+        }
+
         CValidationState state;
         if (!ProcessNewBlockHeaders(headers, state, chainparams, &pindexLast)) {
             int nDoS;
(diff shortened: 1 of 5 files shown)


ouicate:
Please review @patricklodder @chromatic @michilumin

ouicate:
Any update here? @chromatic

src/test/headers_dos_tests.cpp
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2026 The Dogecoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
```
chromatic: Good tests, though I'd rather name this file something like `src/test/headers_validity_tests.cpp`. Even if we're working around potential untrusted client messages, we're testing limits. Not a huge nit though.

src/net_processing.cpp
```diff
@@ -191,6 +191,8 @@ struct CNodeState {
     const CBlockIndex *pindexBestHeaderSent;
     //! Length of current-streak of unconnecting headers announcements
     int nUnconnectingHeaders;
+    //! Low-work side-fork headers accepted from this peer since connect
+    int nLowWorkSideForkHeadersFromPeer;
```
chromatic: Can we start using unsigned ints for these counts? I'm okay diverging from the original Satoshi code; I certainly wouldn't have written network code with signed ints myself.

chromatic:
The `nUnconnectingHeaders` reset-condition change looks like a solid, faithful
port of the upstream fix. I do have two correctness concerns with the new
global/per-peer side-fork accounting that I think need to be resolved before
merge, plus a minor nit.

**1. `nLowWorkSideForkHeaderCount` never decreases at runtime**.

Can this turn into a permanent, self-inflicted outage.  Nothing decrements the
counter while the node is running, so it seems that ordinary network noise may
push this counter toward `MAX_LOW_WORK_SIDEFORK_HEADERS` (2048) over long
uptime. A restart will clear this, but it may penalize honest peers.

Maybe re-derive the count when `pindexBestHeader`/tip advances past
previously-flagged entries, or periodically re-run
`RecountLowWorkSideForkHeaders()`?

**2. Per-peer scoring increments slowly.**

Both good and bad here. `AcceptBlockHeader` increments the *global* counter for
every qualifying header in the message, and a `headers` message can carry up to
`MAX_HEADERS_RESULTS` (2000) headers. So a single peer can exhaust the entire
2048 global budget in one or two messages while their own per-peer counter only
advances by 1 or 2.

Per-peer scoring should count every side-fork header accepted from that peer in
the batch.

You could test this a Python `qa/rpc-tests` test using
  `p2p-acceptblock.py`-style scaffolding: connect a mock P2P peer, send one
  `headers` message containing ~2000 headers off an old ancestor on regtest, then check the node's ban score for
  that peer (`getpeerinfo`'s `banscore`).

**Minor:** `state.DoS(20, error(...), 0, "too-many-sidework-headers")` passes `0`
for the reject code where neighboring calls pass a real `REJECT_*` constant —
intentional, or a copy/paste slip?

**Nit:** the new block re-does `mapBlockIndex.find(block.hashPrevBlock)` to get
`pindexPrev`, but `AcceptBlockHeader` already resolved the identical `pindexPrev`
a few lines earlier. We could reuse that instead of re-deriving it.

ouicate:
Thanks for the review @chromatic — addressed in 4cd7a76.

- Renamed the unit test to `src/test/headers_validity_tests.cpp`.
- `nUnconnectingHeaders`, `nLowWorkSideForkHeadersFromPeer`, and `MAX_UNCONNECTING_HEADERS` are now unsigned.
- The global side-fork count can shrink at runtime. Flagged headers live in a set and are dropped when the tip / best-header advances past them or they later get block data. Recount is O(set), not a full `mapBlockIndex` walk.
- Per-peer scoring now counts every newly accepted side-fork header in the batch (including ones accepted before a later reject in the same message). Added `qa/rpc-tests/p2p-headers-validity.py` to check banscore after a 200-header batch.
- `too-many-sidefork-headers` now uses `REJECT_INVALID`.
- Reuses the already-resolved `pindexPrev` instead of looking it up again.

PR #4020, opened 2026-07-03
depends: fix broken package source downloads

nformant1:
Three depends packages currently fail to fetch because their upstream
source URLs are dead, and the FALLBACK_DOWNLOAD_PATH mirror
(depends.dogecoincore.org) currently serves an expired TLS
certificate, so the fallback fails too:

- openssl: openssl.org no longer serves 1.0.2 under /source/old/1.0.2/
  (404). Still available at the root /source/ path.
  https://openssl-library.org/source/old/1.0.2/index.html

- qrencode: fukuchi.org is offline entirely; upstream now points to
  GitHub for releases instead. See also Bitcoin Core hitting and
  fixing the same issue for qrencode 4.1.1:
  https://repology.org/project/qrencode/information
  https://github.com/bitcoin/bitcoin/pull/33494

- zlib: zlib.net moves previous releases into /fossils once a new
  version is published; root path no longer serves 1.3.
  https://www.zlib.net/fossils/

Each download_path is updated to a source verified to serve a
byte-identical file. No sha256_hash values were changed.

Tested with a clean `make -C depends HOST=x86_64-w64-mingw32` on this
branch, followed by `./autogen.sh && ./configure && make`.

### DIFF
--- a/depends/packages/expat.mk
+++ b/depends/packages/expat.mk
@@ -1,6 +1,6 @@
 package=expat
 $(package)_version=2.6.2
-$(package)_download_path=https://downloads.sourceforge.net/project/expat/expat/$($(package)_version)
+$(package)_download_path=https://github.com/libexpat/libexpat/releases/download/R_2_6_2
 $(package)_file_name=$(package)-$($(package)_version).tar.bz2
 $(package)_sha256_hash=9C7C1B5DCBC3C237C500A8FB1493E14D9582146DD9B42AA8D3FFB856A3B927E0
 
--- a/depends/packages/openssl.mk
+++ b/depends/packages/openssl.mk
@@ -1,7 +1,7 @@
 package=openssl
 $(package)_version=1.0.2
 $(package)_version_suffix=u
-$(package)_download_path=https://www.openssl.org/source
+$(package)_download_path=https://github.com/openssl/openssl/releases/download/OpenSSL_1_0_2u
 $(package)_file_name=$(package)-$($(package)_version)$($(package)_version_suffix).tar.gz
 $(package)_sha256_hash=ecd0c6ffb493dd06707d38b14bb4d8c2288bb7033735606569d8f90f89669d16
 $(package)_patches=secure_getenv.patch
--- a/depends/packages/openssl.mk
+++ b/depends/packages/openssl.mk
@@ -1,7 +1,7 @@
 package=openssl
 $(package)_version=1.0.2
 $(package)_version_suffix=u
-$(package)_download_path=https://www.openssl.org/source/old/$($(package)_version)
+$(package)_download_path=https://www.openssl.org/source
 $(package)_file_name=$(package)-$($(package)_version)$($(package)_version_suffix).tar.gz
 $(package)_sha256_hash=ecd0c6ffb493dd06707d38b14bb4d8c2288bb7033735606569d8f90f89669d16
 $(package)_patches=secure_getenv.patch
--- a/depends/packages/zlib.mk
+++ b/depends/packages/zlib.mk
@@ -1,6 +1,6 @@
 package=zlib
 $(package)_version=1.3
-$(package)_download_path=https://www.zlib.net/fossils
+$(package)_download_path=https://github.com/madler/zlib/releases/download/v$($(package)_version)
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_sha256_hash=ff0ba4c292013dbc27530b3a81e1f9a813cd39de01ca5e0f8bf355702efa593e
 
--- a/depends/packages/zlib.mk
+++ b/depends/packages/zlib.mk
@@ -1,6 +1,6 @@
 package=zlib
 $(package)_version=1.3
-$(package)_download_path=http://www.zlib.net
+$(package)_download_path=https://www.zlib.net/fossils
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_sha256_hash=ff0ba4c292013dbc27530b3a81e1f9a813cd39de01ca5e0f8bf355702efa593e
 
--- a/depends/packages/freetype.mk
+++ b/depends/packages/freetype.mk
@@ -1,6 +1,6 @@
 package=freetype
 $(package)_version=2.11.0
-$(package)_download_path=http://download.savannah.gnu.org/releases/$(package)
+$(package)_download_path=https://download-mirror.savannah.gnu.org/releases/$(package)
 $(package)_file_name=$(package)-$($(package)_version).tar.xz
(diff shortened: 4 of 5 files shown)


nformant1:
Also fixing `expat`, which surfaced in CI but not in my local
x86_64-w64-mingw32 build (expat isn't a dependency for that target):

- expat: downloads.sourceforge.net is unreliable/times out from CI
  runners. Switched to the official GitHub release asset for the
  same version, which serves an identical file (verified against
  the existing sha256_hash). Bitcoin Core's depends system already
  uses this same source for expat:
  https://github.com/bitcoin/bitcoin/blob/master/depends/packages/expat.mk

Tested with a clean native `make -C depends` (no HOST=, i.e.
x86_64-unknown-linux-gnu) on this branch.

nformant1:
Also fixing `freetype`, which showed up as a connection reset in CI
(x86_64-linux-nowallet):

- freetype: download.savannah.gnu.org can reset connections
  mid-transfer. Switched to download-mirror.savannah.gnu.org, which
  the FreeType project explicitly recommends as the fallback for
  exactly this situation:
  https://www.mail-archive.com/freetype@nongnu.org/msg05856.html

  Note: Bitcoin Core's depends still points at the primary host for
  this package, so there's no existing precedent there like there
  was for qrencode/expat — this fix is based on upstream FreeType's
  own guidance plus the reproduced failure in CI. Happy to revert
  this one commit independently if maintainers prefer a different
  approach.

Tested with a clean native `make -C depends` (no HOST=) on this
branch.

depends/packages/openssl.mk
```diff
@@ -1,7 +1,7 @@
 package=openssl
 $(package)_version=1.0.2
 $(package)_version_suffix=u
-$(package)_download_path=https://www.openssl.org/source/old/$($(package)_version)
+$(package)_download_path=https://www.openssl.org/source
```
xanimo:
```suggestion
$(package)_download_path=https://github.com/openssl/openssl/releases/download/OpenSSL_1_0_2u
```
Root `/source/` is a rotating listing — it's the reshuffle that broke the previous path, and it 404'd again in testing (https://github.com/xanimo/dogecoin/actions/runs/28750851620/job/85249574581). The fallback can't catch this package (Bitcoin never shipped 1.0.2u), so the primary carries the whole weight. Release asset is hash-identical; pin unchanged.

depends/packages/zlib.mk
```diff
@@ -1,6 +1,6 @@
 package=zlib
 $(package)_version=1.3
-$(package)_download_path=http://www.zlib.net
+$(package)_download_path=https://www.zlib.net/fossils
```
xanimo:
```suggestion
$(package)_download_path=https://github.com/madler/zlib/releases/download/v$($(package)_version)
```
zlib.net returns HTTP 415 to depends' curl — user-agent filtering, so it verifies in a browser but fails headless (https://github.com/xanimo/dogecoin/actions/runs/28751173056/job/85250459177). Fallback has no zlib 1.3 either. Release asset is the same bytes; pin unchanged.

xanimo (commented):
Concept ACK — thank you for taking this on. Tested the full fix as a stack: your five commits plus three complementary ones, all green end-to-end including macOS: (https://github.com/xanimo/dogecoin/actions/runs/28751371264).
Why this PR alone doesn't pass CI yet: the macOS jobs fetch the SDK from depends.dogecoincore.org (expired TLS cert) via SDK_URL in ci.yml, and FALLBACK_DOWNLOAD_PATH in depends/Makefile points at the same dead host — which is why none of the primary URL failures you fixed here were ever absorbed by the fallback as designed. miniupnpc's direct /files/ path is also gone (free.fr moved to a download.php handler; the existing _download_file mechanism in funcs.mk handles it without touching the hash or cache key).
I hit all three back in June (#4013). They're on (https://github.com/xanimo/dogecoin/tree/refs/heads/fix/ci-sdk-fallback-miniupnpc) as three commits that touch none of your files — cherry-pick them or whatever you want.
Two URL suggestions inline (openssl, zlib), and these come with CI receipts rather than taste: in testing, openssl's root /source/ listing 404'd — the same rotation that broke the previous /source/old/1.0.2/ path — and zlib.net answered depends' curl with HTTP 415: the server filters non-browser user agents, which is why the URL verifies in a browser but fails in any headless build. Critically, the fallback can't catch either package — Bitcoin never shipped openssl 1.0.2u or zlib 1.3, so bitcoincore.org has no copy; both failures went primary-404/415 → fallback-404 → build dead. The GitHub release assets for both are project-published artifacts that don't move when websites reorganize, and both are hash-identical (sha256 pins unchanged, verified by depends during full builds on the June branch). The pattern from the test runs: your GitHub-asset (expat) and immutable-archive (qrencode) choices sailed through; the two project-website URLs are the two that broke. Project websites are maintained for browsers; release artifacts are maintained for machines — and depends is a machine.
Risk class is unchanged throughout: every download, including the SDK (pinned via sdk-shasum in the workflow), remains verified against existing sha256 hashes, so the worst any host can do is fail to serve.
Two housekeeping notes: PR title still says "(openssl, qrencode, zlib)" but the scope is now five packages — suggest dropping the parenthetical. And the dead fallback mirror deserves its own issue regardless of this PR; bitcoincore.org is a working-but-partial stopgap (as the openssl/zlib fallback misses just demonstrated), and a dogecoin-controlled replacement is the real fix.

depends/packages/openssl.mk
nformant1: Thank you! I had already considered the GitHub source, but at first I tried to rely on the most “official” source possible.

chromatic:
I'll look at this over the weekend. I think it's blocking other PRs now.

chromatic:
Looks like boost 1.63 needs a patch for modern GCCs. I'll propose something, perhaps in a separate PR.

depends/packages/expat.mk
```diff
@@ -1,6 +1,6 @@
 package=expat
 $(package)_version=2.6.2
-$(package)_download_path=https://downloads.sourceforge.net/project/expat/expat/$($(package)_version)
+$(package)_download_path=https://github.com/libexpat/libexpat/releases/download/R_2_6_2
 $(package)_file_name=$(package)-$($(package)_version).tar.bz2
 $(package)_sha256_hash=9C7C1B5DCBC3C237C500A8FB1493E14D9582146DD9B42AA8D3FFB856A3B927E0
```
chromatic, line 5: I had to change this to `9c7c1b5dcbc3c237c500a8fb1493e14d9582146dd9b42aa8d3ffb856a3b927e0` (the lower-case version) on my machine. I notice the other hashes in this file match that case scheme.

This makes me wonder if we need to make the file checksum checker case-insensitive; a good first issue outside of this PR perhaps.

chromatic (approved):
After applying #4024 atop this, the build succeeds and all tests pass for me. ACK.

Let's hold this for another couple of days before merging, just in case people want to examine my PR as well.

PR #4022, opened 2026-07-17
Relax RPC auth-scheme validation

chromatic:
RFC 7235 says that the auth-scheme is case-insensitive, so we can allow `Basic` as well as `basic` and any other variation. This improves RPC interoperability slightly.

The helper method should be allocation-free. At some point we may want to reuse it elsewhere. For now it's fine where it is.

### DIFF
--- a/qa/rpc-tests/httpbasics.py
+++ b/qa/rpc-tests/httpbasics.py
@@ -46,7 +46,8 @@ class HTTPBasicsTest (BitcoinTestFramework):
         conn.close()
 
         #same should be if we add keep-alive because this should be the std. behaviour
-        headers = {"Authorization": "Basic " + str_to_b64str(authpair), "Connection": "keep-alive"}
+        #test case-insensitive auth-schema (RFC 7235)
+        headers = {"Authorization": "basic " + str_to_b64str(authpair), "Connection": "keep-alive"}
 
         conn = http.client.HTTPConnection(url.hostname, url.port)
         conn.connect()
@@ -63,7 +64,8 @@ class HTTPBasicsTest (BitcoinTestFramework):
         conn.close()
 
         #now do the same with "Connection: close"
-        headers = {"Authorization": "Basic " + str_to_b64str(authpair), "Connection":"close"}
+        #no one should mix caps like this, but RFC 7235 says they can
+        headers = {"Authorization": "BaSiC " + str_to_b64str(authpair), "Connection":"close"}
 
         conn = http.client.HTTPConnection(url.hostname, url.port)
         conn.connect()
@@ -72,10 +74,20 @@ class HTTPBasicsTest (BitcoinTestFramework):
         assert(b'"error":null' in out1)
         assert(conn.sock==None) #now the connection must be closed after the response
 
+        #a non-Basic scheme must still be rejected
+        headers = {"Authorization": "Bearer " + str_to_b64str(authpair), "Connection":"close"}
+
+        conn = http.client.HTTPConnection(url.hostname, url.port)
+        conn.connect()
+        conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
+        assert_equal(conn.getresponse().status, http.client.UNAUTHORIZED)
+        conn.close()
+
         #node1 (2nd node) is running with disabled keep-alive option
+        #RFC 7235 allows this monstrosity
         urlNode1 = urllib.parse.urlparse(self.nodes[1].url)
         authpair = urlNode1.username + ':' + urlNode1.password
-        headers = {"Authorization": "Basic " + str_to_b64str(authpair)}
+        headers = {"Authorization": "bASIC " + str_to_b64str(authpair)}
 
         conn = http.client.HTTPConnection(urlNode1.hostname, urlNode1.port)
         conn.connect()
--- a/src/utilstrencodings.h
+++ b/src/utilstrencodings.h
@@ -1,5 +1,6 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
 // Copyright (c) 2009-2016 The Bitcoin Core developers
+// Copyright (c) 2026 The Dogecoin Core developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -143,5 +144,20 @@ bool TimingResistantEqual(const T& a, const T& b)
  * @note The result must be in the range (-10^18,10^18), otherwise an overflow error will trigger.
  */
 bool ParseFixedPoint(const std::string &val, int decimals, int64_t *amount_out);
+
+/**
+ * Converts the given character to its lowercase equivalent.
+ * This function is locale independent. It only converts uppercase
+ * characters in the standard 7-bit ASCII range.
+ * This is a feature, not a limitation.
+ *
+ * @param[in] c     the character to convert to lowercase.
+ * @return          the lowercase equivalent of c; or the argument
+ *                  if no conversion is possible.
+ */
+constexpr unsigned char ToLower(unsigned char c)
+{
+    return (c >= 'A' && c <= 'Z' ? (c - 'A') + 'a' : c);
+}
 
 #endif // BITCOIN_UTILSTRENCODINGS_H
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -1,4 +1,5 @@
 // Copyright (c) 2015-2016 The Bitcoin Core developers
+// Copyright (c) 2026 The Dogecoin Core developers
 // Distributed under the MIT software license, see the accompanying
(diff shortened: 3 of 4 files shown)


michilumin (approved):
..... Compliance! <3 uh. Someone's gotta do it!

xanimo:
Agreed on the intent — RFC 7235 is clear that auth-scheme is case-insensitive, and the equivalence rewrite looks right to me (the `size() < 6` guard matches the old `substr(0, 6)` behavior for short inputs, and `substr(6)` + `boost::trim` still handles `1*SP` as before).

One thing on the folding primitive, though: `std::tolower` is locale-dependent, and the Turkish-I case breaks the comparison:

```
C            : tolower('I')=0x69  tolower('i')=0x69  match=yes
tr_TR.UTF-8  : tolower('I')=0x49  tolower('i')=0x69  match=NO
tr_TR        : tolower('I')=0xfd  tolower('i')=0x69  match=NO
en_US.UTF-8  : tolower('I')=0x69  tolower('i')=0x69  match=yes
```

So under a Turkish locale `case_insensitive_equal("BASIC", "basic", 5)` returns false and `Authorization: BASIC …` gets rejected — the interop this PR is adding. Here's the new check's behaviour across locales:

```
                 C locale        tr_TR.UTF-8
"basic x"        accepted        accepted
"BaSiC x"        accepted        accepted
"bASIC x"        accepted        REJECTED
"BASIC x"        accepted        REJECTED
```

(Inputs with no case variation — `""`, `"Basic"`, `"Basic x"`, `"Basicx"`, `" Basic x"` — behave identically to the old check under both locales.)

That includes the new `"bASIC "` vector at `httpbasics.py:81`.

Reachability is narrow: `setlocale` doesn't appear anywhere in `src/`, so `dogecoind` stays in the `C` locale and is fine today. But `dogecoin-qt` with `-server` reaches the same RPC init, and per Qt's docs "On Unix/Linux Qt is configured to use the system locale settings by default" (they suggest calling `setlocale(LC_NUMERIC,"C")` after constructing the app to undo part of that). Worth saying that this can only cause false rejects, never false accepts, so it's a correctness wart rather than anything security-relevant.

A locale-independent ASCII fold sidesteps it and stays allocation-free:

```cpp
static bool case_insensitive_equal(const char* a, const char* b, std::size_t n)
{
    for (std::size_t i = 0; i < n; ++i) {
        unsigned char ac = static_cast<unsigned char>(a[i]);
        unsigned char bc = static_cast<unsigned char>(b[i]);
        if (ac >= 'A' && ac <= 'Z') ac += 'a' - 'A';
        if (bc >= 'A' && bc <= 'Z') bc += 'a' - 'A';
        if (ac != bc) return false;
    }
    return true;
}
```

This is why Bitcoin Core carries its own `ToLower` instead of using `std::tolower`. There's no equivalent helper in this tree yet, so rolling it by hand here seems right.

One small thing while you're in there: `inline` gives the helper external linkage under a fairly generic name, while its neighbours (`multiUserAuthorized`, `RPCAuthorized`) are `static`. Since you mention possibly reusing it later, `static` for now might be tidier.

Unrelated to your changes: the red CI here is the pre-existing depends breakage (`openssl-1.0.2u.tar.gz` 404s), which #4020 fixes — not anything in this PR. Side effect is that this code hasn't actually been compiled by CI yet.

chromatic:
I've folded in Bitcoin's `ToLower` to address this case. I'm not sure how best to test the Turkish locale's rules without more difficult surgery on launching the RPC node.

src/utilstrencodings.h
```diff
@@ -144,4 +145,19 @@ bool TimingResistantEqual(const T& a, const T& b)
...
+ * Converts the given character to its lowercase equivalent.
+ * This function is locale independent. It only converts uppercase
+ * characters in the standard 7-bit ASCII range.
+ * This is a feature, not a limitation.
+ *
+ * @param[in] c     the character to convert to lowercase.
+ * @return          the lowercase equivalent of c; or the argument
+ *                  if no conversion is possible.
+ */
+constexpr char ToLower(char c)
```
xanimo: Match Bitcoin Core's signature (`unsigned char`) to avoid an overload collision on the next util sync, and to drop the implicit conversion (`case_insensitive_equal` already passes `unsigned char`):

```suggestion
constexpr unsigned char ToLower(unsigned char c)
```

qa/rpc-tests/httpbasics.py
```diff
@@ -73,9 +75,10 @@ def run_test(self):
         assert(conn.sock==None) #now the connection must be closed after the response
```
xanimo, line 75: Add a negative case so the scheme check is proven to still *reject* — valid credentials but a non-`basic` scheme must 401:

```suggestion
        assert(conn.sock==None) #now the connection must be closed after the response

        #a non-Basic scheme must still be rejected
        headers = {"Authorization": "Bearer " + str_to_b64str(authpair), "Connection":"close"}
        conn = http.client.HTTPConnection(url.hostname, url.port)
        conn.connect()
        conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
        assert_equal(conn.getresponse().status, 401)
        conn.close()
```

xanimo (commented):
Built at `55f9ea4d80` (depends toolchain): clean build, `make check` green, `httpbasics.py` passes. LGTM with a few notes.

**1. Match Bitcoin Core's `ToLower` signature.** Bitcoin Core already defines this as `constexpr unsigned char ToLower(unsigned char c)`. A `char` variant with the same name collides as an overload whenever util is next synced from upstream. It's also cleaner here, since `case_insensitive_equal` already passes `unsigned char`. (inline suggestion below)

**2. Assert the reject path.** The test broadens acceptance but never checks that a bad scheme is still rejected — cheap to lock in. (inline suggestion below)

**3. Testing the Turkish-I rule** — no need to run the node under `tr_TR`. The hazard is that `std::tolower` is locale-dependent:

```
C / en_US.UTF-8 : tolower('I') = 0x69 ('i')   -> "BASIC" folds to "basic"
tr_TR.UTF-8     : tolower('I') = 0x49 ('I')   -> stays uppercase, match fails
tr_TR           : tolower('I') = 0xfd ('ı')   -> dotless i, match fails
```

`ToLower` avoids this by construction: it's `constexpr`, so `ToLower('I')` is computed at compile time as `('I'-'A')+'a'` = `0x49-0x41+0x61` = `0x69` ('i'), with no locale consulted, ever. So a `static_assert` *is* the proof — it holds at compile time and needs no locale installed. Add to `src/test/util_tests.cpp` if you want:

```cpp
BOOST_AUTO_TEST_CASE(util_ToLower)
{
    // Locale-independent ASCII fold. std::tolower() is not: tolower('I') is
    // 0x49 under tr_TR.UTF-8 (stays 'I') and 0xfd under tr_TR, not 0x69 ('i'),
    // so "BASIC" would fail to match "basic". ToLower is constexpr and cannot
    // consult a runtime locale, so ToLower('I') is always 0x69.
    static_assert(ToLower('A') == 'a', "");
    static_assert(ToLower('I') == 'i', "the Turkish-I case");
    static_assert(ToLower('Z') == 'z', "");
    static_assert(ToLower('a') == 'a', "");
    static_assert(ToLower('i') == 'i', "");
    static_assert(ToLower('5') == '5', "non-letters are unchanged");
    BOOST_CHECK(ToLower('I') == 'i');
}
```

chromatic:
Thanks! I've pushed another commit that addresses your suggestions, @xanimo . If this looks good, I'll squash things and request a final approval.

xanimo (commented):
ACK faf830a — both suggestions applied, and ToLower matches upstream
exactly, so the next util sync will conflict cleanly instead of quietly
adding an overload.

Checked the rewrite against the old one across the whole first six bytes:
same accept set apart from the case variants of "basic ", which is the
point. Nothing else moved, and the fold is locale-proof now.

One thing I should have caught in my own suggestion: 115 and 121 use
http.client.NOT_FOUND / BAD_REQUEST, so UNAUTHORIZED would be more
consistent than the bare 401 I handed you. Also util_tests.cpp didn't pick
up the Dogecoin copyright line the other two files got.

Squash away and I'll approve.

chromatic:
I've squashed the two commits into one and changed the constant `401` to the appropriate named constant.

xanimo (approved):
tested ACK 7effe1013c builds clean; util_ToLower + util_tests green; httpbasics.py passes (basic/BaSiC/bASIC accepted, Bearer→unauth). RFC 7235 rationale and locale-independent ToLower both check out.

michilumin (approved):
re-review; good expansions. bouncing back to chromatic for more eyes.

chromatic:
I force-pushed to remove the stale extra commit. @michilumin one final review please.

xanimo (approved):
reACK f82142387f, parity with 7effe1013

michilumin (approved):
thorough, at least

PR #4024, opened 2026-07-18
Fix `make -C depends` builds with GCC 15

chromatic:
This version of GCC, shipped with newer Linux releases, is stricter so we need to patch some dependencies and relax some new compiler options for others.

The approach here should preserve the same behavior.

This PR applies cleanly atop #4020 ; we should merge that first.

### DIFF
--- a/depends/packages/bdb.mk
+++ b/depends/packages/bdb.mk
@@ -9,7 +9,9 @@ define $(package)_set_vars
 $(package)_config_opts=--disable-shared --enable-cxx --disable-replication
 $(package)_config_opts_mingw32=--enable-mingw
 $(package)_config_opts_linux=--with-pic
+$(package)_config_opts_linux+=--with-mutex=POSIX/pthreads/library
 $(package)_cxxflags=-std=c++11
+$(package)_cflags+=-Wno-incompatible-pointer-types
 endef
 
 define $(package)_preprocess_cmds
--- a/depends/packages/boost.mk
+++ b/depends/packages/boost.mk
@@ -3,6 +3,7 @@ $(package)_version=1_63_0
 $(package)_download_path=https://sourceforge.net/projects/boost/files/boost/1.63.0
 $(package)_file_name=$(package)_$($(package)_version).tar.bz2
 $(package)_sha256_hash=beae2529f759f6b3bf3f4969a19c2e9d6f0c503edcb2de4a61d1428519fcb3b0
+$(package)_patches=fix-boost-1.63-missing-headers.patch
 
 define $(package)_set_vars
 $(package)_config_opts_release=variant=release
@@ -25,7 +26,8 @@ $(package)_cxxflags_linux=-fPIC
 endef
 
 define $(package)_preprocess_cmds
-  echo "using $(boost_toolset_$(host_os)) : : $($(package)_cxx) : <cxxflags>\"$($(package)_cxxflags) $($(package)_cppflags)\" <linkflags>\"$($(package)_ldflags)\" <archiver>\"$(boost_archiver_$(host_os))\" <striper>\"$(host_STRIP)\"  <ranlib>\"$(host_RANLIB)\" <rc>\"$(host_WINDRES)\" : ;" > user-config.jam
+	echo "using $(boost_toolset_$(host_os)) : : $($(package)_cxx) : <cxxflags>\"$($(package)_cxxflags) $($(package)_cppflags)\" <linkflags>\"$($(package)_ldflags)\" <archiver>\"$(boost_archiver_$(host_os))\" <striper>\"$(host_STRIP)\"  <ranlib>\"$(host_RANLIB)\" <rc>\"$(host_WINDRES)\" : ;" > user-config.jam && \
+	patch -p1 < $($(package)_patch_dir)/fix-boost-1.63-missing-headers.patch
 endef
 
 define $(package)_config_cmds
--- a/depends/packages/expat.mk
+++ b/depends/packages/expat.mk
@@ -2,7 +2,7 @@ package=expat
 $(package)_version=2.6.2
 $(package)_download_path=https://github.com/libexpat/libexpat/releases/download/R_2_6_2
 $(package)_file_name=$(package)-$($(package)_version).tar.bz2
-$(package)_sha256_hash=9C7C1B5DCBC3C237C500A8FB1493E14D9582146DD9B42AA8D3FFB856A3B927E0
+$(package)_sha256_hash=9c7c1b5dcbc3c237c500a8fb1493e14d9582146dd9b42aa8d3ffb856a3b927e0
 
 define $(package)_config_cmds
   $($(package)_autoconf)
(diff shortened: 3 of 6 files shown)


xanimo:
Thanks for tackling this — the GCC 15 / Python 3.12 breakage is going to bite more people as newer distros land.

One blocker: depends/patches/boost/fix-boost-1.63-missing-headers.patch isn't in the PR. boost.mk declares it via $(package)_patches and applies it in preprocess_cmds, but the changed-file list is just the five .mk files, and depends/patches/boost/ doesn't exist in the tree. Looks like a missed git add.

Worth being explicit that this is the current CI failure rather than the pre-existing broken downloads — all 11 build jobs die here:

```
sha256sum: .../depends/patches/boost/fix-boost-1.63-missing-headers.patch: No such file or directory
/bin/sh: 1: cd: can't cd to .../depends/patches/boost
cp: cannot stat 'fix-boost-1.63-missing-headers.patch': No such file or directory
make: *** [funcs.mk:247: .../boost/1_63_0-1489e0e8121/.stamp_preprocessed] Error 1
```

chromatic:
That's what I get for not triple-checking my PR or CI. Thanks; I've added the actual patch now and force-pushed.

WXZttnn:
How can I get my payment?

xanimo (commented):
Tested this properly now. Native Linux is good; the Windows target is broken by the bdb change.

## Native `x86_64-pc-linux-gnu` — all good

Full depends rebuild under GCC 15.3.0 (Debian 15.3.0-1) in a container — every package artifact came from this run, nothing reused from a prior build. All five patched packages build through:

| package | build_id |
|---|---|
| bdb | `b5028513eaf` |
| boost | `c33ae5a682c` |
| expat | `e4b36fec114` |
| qt | `a7f0b002233` |
| xcb_proto | `357d206a883` |

Confirmed the objects really came from 15.3 rather than a cached earlier toolchain — `GCC: (Debian 15.3.0-1) 15.3.0` is stamped into `libdb-5.3.a`, `libboost_system-mt.a` and `libQt5Core.a`. The boost patch applies cleanly (on both the linux and mingw32 hosts), the `imp` → `importlib.util` rewrite works, and `-Wno-incompatible-pointer-types` is sufficient for bdb — no `-Wno-error=` form needed.

## `x86_64-w64-mingw32` — bdb fails

```
../dist/./../src/dbinc/mutex_int.h:1012:31: error: 'DB_MUTEX' {aka 'struct __db_mutex_t'}
    has no member named 'sharecount'
 1012 |         (atomic_read(&(mutexp)->sharecount) != 0) :             \
make[1]: *** [Makefile:2201: mut_pthread.o] Error 1
```

This is `$(package)_config_opts+=--with-mutex=POSIX/pthreads/library` landing on the unconditional opts and reaching the Windows target.

BDB picks Win32 mutexes on mingw by default — `dist/aclocal/mutex.m4`:

```sh
# User-specified Win32 mutexes (MinGW build)
if test "$db_cv_mingw" = yes; then
	db_cv_mutex=win32/gcc
```

and the two modes are mutually exclusive by construction — `src/dbinc/mutex_int.h`:

```c
#if defined(HAVE_MUTEX_HYBRID) || defined(DB_WIN32) || \
	(defined(HAVE_SHARED_LATCHES) && !defined(HAVE_MUTEX_PTHREADS))
#define MUTEXP_IS_BUSY(mutexp)                    \
	(F_ISSET(mutexp, DB_MUTEX_SHARED) ?           \
	(atomic_read(&(mutexp)->sharecount) != 0) :   \
	F_ISSET(mutexp, DB_MUTEX_LOCKED))
```

`DB_WIN32` is defined for the Windows target, so `MUTEXP_IS_BUSY` takes the `sharecount` branch — but `sharecount` only exists in the non-pthreads variant of the struct ("Pthreads shared latches use pthread_rwlock instead"). Forcing `HAVE_MUTEX_PTHREADS` on Windows selects a macro referencing a field that isn't there.

**Fix:** scope it to Linux, mirroring the `--with-pic` line right below it:

```make
$(package)_config_opts_linux+=--with-mutex=POSIX/pthreads/library
```

To be clear the option itself is right, and I checked all four combinations rather than assuming:

| | with `--with-mutex=POSIX/pthreads/library` | without |
|---|---|---|
| `x86_64-pc-linux-gnu` | builds | `configure: error: Support for FCNTL mutexes was removed in BDB 4.8.` |
| `x86_64-w64-mingw32` | `error: 'DB_MUTEX' has no member named 'sharecount'` | builds, `checking for mutexes... win32/gcc` |

So the option is genuinely needed on Linux — BDB 5.3.28's mutex autodetection no longer works under GCC 15 (its probes fail to compile), so it falls back to `UNIX/fcntl` and then rejects its own fallback. And it's genuinely harmful on Windows, where dropping it lets BDB pick `win32/gcc` by itself and the build completes.

`_config_opts_linux` is the placement that satisfies both.

## Unrelated, but you'll hit it

On a modern toolchain the mingw depends build also fails earlier, at libevent:

```
/usr/share/mingw-w64/include/iphlpapi.h:40:30: error: unknown type name 'PMIB_TCPTABLE2'
```

Nothing sets `_WIN32_WINNT` in `hosts/mingw32.mk`, so newer mingw-w64 headers declare Vista+ APIs whose types stay behind version guards. Not caused by this PR — it doesn't touch libevent — and CI hasn't surfaced it because the win jobs die earlier, on the pre-existing `openssl-1.0.2u.tar.gz` 404 that #4020 fixes. Whether the `ubuntu:20.04` toolchain hits it at all I can't say without a run that gets that far. I only mention it because I had to build bdb in isolation to get around it; probably its own PR if anyone cares about cross-building on a current distro.

michilumin (approved):
Extremely straightforward requirement bump that allows us to tighten up deps going forward. 
Want to make sure to address Bluezr's mentions above with mingw, since we do need to keep compatibility.

chromatic:
I'll review the mingw after we get CI working. Your diagnosis makes sense.

chromatic:
I've rebased this after the #4020 merge and will do the same after the #4026 merge. I expect to see the mingw failures; if CI shows them, I have a clear path to prove that they're fixed.

chromatic:
I've rebased and will monitor CI now that #4026 has merged. I expect any errors will be obvious here.

michilumin (approved):
small change, ack lgtm

xanimo (approved):
ACK. Tested d1ff6e1. make -C depends boost passes cleanly with GNU patch (Linux default). Note for macOS contributors: Apple's BSD patch 2.0 misparses the bare blank line between the two @@ hunks in the filesys.h section and fails non-interactively — gpatch from Homebrew is required.

Nit on bdb.mk: _config_opts_linux+= scopes the mutex flag to Linux, but the _cflags+= on the next line lands in the base variable that every host inherits (funcs.mk:11). Since _cflags_linux+= is already handled by funcs.mk:100, it would scope the warning flag to match:

$(package)_config_opts_linux+=--with-mutex=POSIX/pthreads/library
$(package)_cflags_linux+=-Wno-incompatible-pointer-types
Only suppresses a diagnostic either way, so not a blocker — but on mingw32 BDB takes the Win32 mutex path, so the current form silences something that was never deliberately looked at.

Minor note on the boost patch: zero-context hunks place by offset only — fine with the sha256 pin, but worth regenerating with context if the version ever bumps.

Issue #4003, opened 2026-05-23
[feat]  Implement support for Qt6 dependencies


victorsk2019:
# Feature Request

Hi,

As I understand, Qt5 is being masked for removal in Gentoo Linux and so I was wondering how much of a task it would be to implement Qt6 support in Dogecoin Core?  Per #3461 seems like adding Qt6 support is a fairly involved process though.  However, without Qt5 packages or them gradually becoming deprecated, installing Dogecoin Core with Qt gui selection may be a somewhat problematic on some systems.       

I recently tried to port net-p2p/dogecoin-qt package to Qt6 in Gentoo ebuild using equivalent Qt5 module dependencies that are being used for Dogecoin Core like so:

`>=dev-qt/qtbase-6.10.1:6[concurrent,dbus,gui,network,widgets,X]`
`>=dev-qt/qttools-6.10.1:6[linguist]`

and that didn't work because Qt5 dependency requirement is part of Dogecoin Core configure process with Qt5 being the latest version supported which seems to be tightly coupled with other Dogecoin Core parts:

`checking for Qt5Core Qt5Gui Qt5Network Qt5Widgets Qt5PrintSupport... no`
`configure: WARNING: Qt dependencies not found; dogecoin-qt frontend will not be built`
`checking whether to build Dogecoin Core GUI... no (Qt5)`
 
**Describe Preferred Solution**

Implement Qt6 support used for gui selection into Dogecoin Core.  From working on Qt6 module dependencies I used for Gentoo ebuilds (i.e. concurrent,dbus,gui,network,widgets,X, linguist), I assume we would need the same modules for Dogecoin Core and since there is already equivalent Qt5 modules that are currently be used to build Dogecoin Core, I would suggest looking into integrating the same but newer Qt6 modules.  Thank you!

victorsk2019:
Based on what I've been reading, it seems the best approach would be to port to version 5.15.0 first.

PR #4027, opened 2026-07-22
Implementing Qt6 support since Qt5 is gradually becoming obsolete 

victorsk2019:
Hi everyone,

This draft proposal aims to address issue #4003.  Some distributions like Gentoo-based Calculate Linux are now using Qt6 with Qt5 being marked for removal on Gentoo.  My plan was to study the config files and iteratively changes with each PR until moving to implementing Qt6 API changes that were introduced in the new Qt version.  However, I decided to give AI (Copilot) a try and it modified the m4 and  configure scripts for me.  To resolve compilation errors due to namespace and function name mismatches I also had to make use of Google's Gemini because this issue is a bit urgent right now and if there is are available tools that can expedite the process I, might as well use it.  

This PR compiles successfully under GCC 15.3.0 and with Qt6 6.11.1, using configure option `--with-gui=qt6`.  The `dogecoin-qt` binary launches Dogecoin Core GUI client but it will need a lot of testing, of course.  

Thank you,
Victor.


chromatic:
I'm glad to see this underway. I've long had it on my list to see when/if we need to move off of the old version of Qt 5 we have, so this will help us plan more work.

PR #4040, opened 2026-08-06
depends: update libevent to 2.1.13-stable

xanimo:
`depends` pins libevent **2.1.12-stable (2020-07-05)**. **2.1.13-stable** was released **2026-07-01** carrying seven security advisories. Four are in `evhttp`, which the RPC server is built on.

| Advisory | Issue |
|---|---|
| GHSA-2gmv-p5m7-98p6 | HTTP trailers permitted header smuggling |
| GHSA-q39v-w2g7-gr8j | header parsing permitted request smuggling |
| GHSA-jcwh-pvf2-73p2 | CRLF and NUL handling allowed parser mismatch |
| GHSA-cvq5-vrvr-j338 | AF_UNIX heap out-of-bound write (`-DNDEBUG` builds) |
| GHSA-c2pj-cg4r-88c8 | dangling pointer in `evbuffer_add_reference` |
| GHSA-58rx-7448-jw47 | OOB write in evdns `dnsname_to_labels` |
| GHSA-fj29-64w6-73h6, GHSA-45c6-qx49-89m8 | evtag/evrpc — **not used by this tree** |

## Reachability

Worth being precise, because it decides how much the parser bugs matter. libevent parses the request line, headers and body **before** handing anything to the registered callback:

```
httpserver.cpp:427   evhttp_set_gencb(http, http_request_cb, NULL)
httpserver.cpp:251   http_request_cb()
httprpc.cpp:150      HTTPReq_JSONRPC()
httprpc.cpp:158      req->GetHeader("authorization")
httprpc.cpp:166      RPCAuthorized()
```

Authentication is step four. Everything libevent parses, it parses before that — so the three parser advisories are reachable by **anyone who can open a connection to the RPC port, with no credentials**.

That isn't critical standing alone: RPC binds localhost and requires auth for any action. But it removes the "authenticated attacker only" mitigation, and request smuggling is specifically an attack against a proxy in front of a backend — a common way RPC gets exposed.

## `glibc_compatibility.patch` had to be regenerated

It **fails** against 2.1.13. The cause is cosmetic — a newer autoconf writes `'getrandom'` where the old one wrote `` `getrandom' `` — so the context lines no longer match. The hunk is otherwise identical, at the same offset, removing the same two lines.

**This is the important part to verify, because a silently-skipped patch is exactly how `getrandom@GLIBC_2.25` would sneak in and fail symbol-check.** Applied against a pristine 2.1.13 tree:

```
[...]

PR #4045, opened 2026-08-12
guix: port the release build from gitian

xanimo:
Adds `contrib/guix`, a Guix-based release build covering every host the gitian
descriptors build: x86_64 / i686 / armhf / aarch64 linux, i686 and x86_64 mingw,
and x86_64 macOS. Pinned to Guix v1.4.0 (`8e2f32ce`). Ported from this tree's own
`contrib/gitian-descriptors/*.yml` rather than from Bitcoin Core's target set, so
the artifact names and contents match what we ship today.

All seven hosts build from one invocation, and all fourteen comparable artifacts
have file lists identical to gitian's: the four linux tarballs, osx64,
osx-unsigned, win32, win64 and every debug archive. glibc floors sit under the
`symbol-check.py` ceilings: 2.10 against a 2.11 limit, and aarch64 at 2.17,
which is that architecture's own floor. The macOS dmg was shown byte-identical
across two independent runs, `cmp` clean over all 30,779,392 bytes, once
genisoimage was taught to honour `SOURCE_DATE_EPOCH`. That reproducibility check
has only been done for darwin, on one machine.

gitian is untouched and still works. The `depends/` commits are the only ones
that could affect it, so gitian's windows descriptor was run against this
branch's depends changes and produced all seven windows artifacts; those
changes are demonstrated inert there rather than merely argued. The darwin
changes are guarded by `FORCE_USE_SYSTEM_CLANG`, which gitian never sets; CI's
macOS job exercises the other side of those guards.


chromatic:
Is the intent to run guix and gitian in parallel for a while, then drop gitian if we find guix better along one or multiple dimensions? I can agree with that in principle as an experiment, though my concern is the effort to maintain two systems in parallel.

In practice we haven't spent a *lot* of effort maintaining gitian lately however.

xanimo:
yes. both for one release cycle, compare, drop gitian if guix holds, meaning all seven hosts byte-identical across two machines. as the description says, that check is darwin only so far, on one machine.

on maintenance you answered that on #4040: three descriptors, seven hosts, today. and gitian-builder has had six commits in 2024, two in 2025, none since april. it still works but isn't where the toolchain work happens anymore.

PR #4046, opened 2026-08-12
ci: add push and pull request branch triggers

edtubbs:


chromatic:
I'm open to this but I'm curious what motivates this; are the CI filters now too strict for specific workstreams?

edtubbs:
> I'm open to this but I'm curious what motivates this; are the CI filters now too strict for specific workstreams?

We were attempting to figure out why actions weren't triggering. Seems it was github related so we're probably good with the current set and changes aren't needed.

chromatic:
In that case, should we close this now and re-open if the issue reoccurs?

PR #4050, opened 2026-08-29
Improve startup by skipping segwit validation

chromatic:
Dogecoin inherited but does not use SegWit code from Bitcoin. Because Dogecoin does not have SegWit enabled, the SegWit validation check on startup is an expensive no-op.

Disabling it for now, at least until we enable SegWit, is a cheap but effective startup performance optimization.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -3858,6 +3858,16 @@ bool RewindBlockIndex(const CChainParams& params)
 {
     LOCK(cs_main);
 
+    // Dogecoin: SegWit is permanently disabled (see IsWitnessEnabled above),
+    // so every branch below - which exists only to re-validate blocks whose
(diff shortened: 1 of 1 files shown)


Jennie867-5309:
What the hell


Jennifer Wages

Jennie867-5309:
Cheap says it all


Jennifer Wages

xanimo (approved):
ACK tested 4b1657a402 on x86_64-pc-linux-gnu

one thing to flag for whenever segwit does get enabled: the guard tests
chainActive.Tip(), which is sound for the first loop since versionbits is
monotone along a chain, but the closing scan tests
IsWitnessEnabled(pindexIter->pprev) on blocks explicitly not in chainActive, and
a side branch computes its own versionbits state, so the tip's state does not
imply it. unreachable while IsWitnessEnabled is return false.

PR #4052, opened 2026-08-30
Document index DB integration and RPC commands

Tsukimarf:
Added documentation for integrating the index DB with new RPC commands and installation instructions.


chromatic:
I think this also may be code for a fork accidentally targeting the Dogecoin Core upstream repository. I'm also closing this one, but feel free to re-open if I misunderstood and you're proposing a change to Dogecoin Core itself.

PR #4055, opened 2026-09-07
Improve dependency fetching resilience

chromatic:
- increase number of retries
- catch and recover from SSL/TLS resets

The intent is to recover from network flakiness in CI and reduce the risk of falling back to the fallback URLs, which have themselves had flakiness in June
- September 2026.


xanimo (commented):
ci runs in ubuntu:20.04 containers which ship curl 7.68.0, and --retry-all-errors needs 7.71.0, so the unknown option kills both the primary and the mirror fetch. probing keeps the flag wherever curl has it:
```
--- a/depends/Makefile
+++ b/depends/Makefile
@@ -16,6 +16,8 @@ BASEDIR = $(CURDIR)
 HASH_LENGTH:=11
 DOWNLOAD_CONNECT_TIMEOUT:=10
 DOWNLOAD_RETRIES:=5
+# --retry-all-errors needs curl >= 7.71.0; CI containers ship focal's 7.68.0
+CURL_RETRY_ALL_ERRORS := $(shell curl --help all 2>/dev/null | grep -q -- --retry-all-errors && echo --retry-all-errors)
 HOST_ID_SALT ?= salt
 BUILD_ID_SALT ?= salt
 
--- a/depends/builders/darwin.mk
+++ b/depends/builders/darwin.mk
@@ -7,7 +7,7 @@ build_darwin_OTOOL: = $(shell xcrun -f otool)
 build_darwin_NM: = $(shell xcrun -f nm)
 build_darwin_INSTALL_NAME_TOOL:=$(shell xcrun -f install_name_tool)
 build_darwin_SHA256SUM = shasum -a 256
-build_darwin_DOWNLOAD = curl --location --fail --connect-timeout $(DOWNLOAD_CONNECT_TIMEOUT) --retry $(DOWNLOAD_RETRIES) --retry-all-errors --retry-max-time 60 -o
+build_darwin_DOWNLOAD = curl --location --fail --connect-timeout $(DOWNLOAD_CONNECT_TIMEOUT) --retry $(DOWNLOAD_RETRIES) $(CURL_RETRY_ALL_ERRORS) --retry-max-time 60 -o
 
 #darwin host on darwin builder. overrides darwin host preferences.
 darwin_CC=$(shell xcrun -f clang) -mmacosx-version-min=$(OSX_MIN_VERSION)
--- a/depends/builders/linux.mk
+++ b/depends/builders/linux.mk
@@ -1,2 +1,2 @@
 build_linux_SHA256SUM = sha256sum
-build_linux_DOWNLOAD = curl --location --fail --connect-timeout $(DOWNLOAD_CONNECT_TIMEOUT) --retry $(DOWNLOAD_RETRIES) --retry-all-errors --retry-max-time 60 -o
+build_linux_DOWNLOAD = curl --location --fail --connect-timeout $(DOWNLOAD_CONNECT_TIMEOUT) --retry $(DOWNLOAD_RETRIES) $(CURL_RETRY_ALL_ERRORS) --retry-max-time 60 -o
```
verified in the ubuntu:20.04 image: the probe expands empty and the fetch of ccache-3.3.3.tar.bz2 matches the pinned sha256.

xanimo (commented):
two additions to the above.

the probe has to be `curl --help all`, not `curl --help`. on 7.68.0 `--help all` exits 0 and prints the general listing, so it degrades correctly there. but modern curl omits `--retry-all-errors` from the plain `--help` output, so a bare `--help` probe returns a false negative on every current system and quietly drops the flag everywhere.

and worth surfacing the degraded case, otherwise an empty probe silently restores the old behavior:

```make
ifeq ($(CURL_RETRY_ALL_ERRORS),)
$(warning depends: curl lacks --retry-all-errors, some errors including tls resets will not be retried)
endif
```

`$(warning)` rather than `$(info)` so it lands on stderr.

xanimo (changes requested):
status=$? sits after fi; a false if with no else exits 0, so total failure exits 0 and the log prints "exit 0". verified: a 404 with max attempts 2 exits 0.

fix:

    curl --location --fail --connect-timeout "$connect_timeout" -o "$outfile" "$url"
    status=$?
    if [ "$status" -eq 0 ]; then
        exit 0
    fi

chromatic:
Good point; don't want to catch curl's 404 errors and drop them. How's this?

xanimo (approved):
looks good to me. tACK
```
$ bash download-retry.sh 2 2 out.bin https://github.com/definitely/not/a/real/path/x.tar.gz
curl failed (exit 22) on attempt 1/2, retrying in 1s...
$ echo $?
22
```

fedikan:
Tested the download wrapper and `fetch_file` caller at `3e55659e53bc8272e7d1a6719224b6652585b797` with an offline fixture matrix: **20/20 checks passed** on Linux (Bash 5.2.21, GNU Make 4.3).

Reproducible harness, commands and per-case results: https://github.com/machine-of-earn/dogecoin-download-review

Coverage beyond the already-reviewed exit-status fix:

- Simulated curl exit 35 followed by success retries once and avoids the mirror. Five failures return 35 after five calls and sleeps of 1, 2, 4 and 8 seconds. Mixed failures preserve the last status; a one-attempt call does not sleep.
- The output filename and URL remain single arguments with spaces/query parameters. A shorter successful result replaces the partial file; a separate real-curl `file://` check confirms the overwrite behavior.
- Using the unmodified `depends/funcs.mk` and Linux builder expression: primary exhaustion reaches the mirror, including the different primary and mirror archive names; primary checksum failure also reaches the mirror. Failure at both locations, or two checksum mismatches, leaves no final source archive. Successful recovery verifies the hash and removes the temporary download directory. An existing cached source skips fetching.

One behavior to keep explicit: `DOWNLOAD_RETRIES=5` now means five total attempts per URL, and exit 22 is retried too. That is consistent with this wrapper; the fixture records it rather than assuming curl's old `--retry` semantics.

The Linux and Darwin download expressions are identical, but this does **not** cover execution on macOS, real TLS resets, curl 7.68, a complete dependency build, or the node test suites. Curl and sleep are stubbed except for the local-file overwrite check. No new defect found in this bounded review. The earlier status fix and old-curl compatibility work belong to xanimo's review above.

PR #4058, opened 2026-09-10
rpc: add createauxblocklight / submitauxblocklight for Stratum use

ab-emcd:
**Motivation**

  Stratum needs control over the aux-chain coinbase vout (per-pool/per-miner payout splits
  decided at submit time), which createauxblock/submitauxblock can't offer since they fix the coinbase at
  template time. getblocktemplate would work but pushes AuxPoW-specific concerns (chain ID, merge-mining
  version bit, target formatting) onto the client — we want the node to keep owning those, just not the
  coinbase.

**What changed**

  - createauxblocklight — caches a template's invariant (non-coinbase) parts under a job_id; returns
    header fields, coinbasevalue, and a merkle branch.
  - submitauxblocklight <jobid> <coinbase> <auxpow> — rebuilds the block from the cached job +
    client-supplied coinbase, recomputes the merkle root, attaches the AuxPoW, submits via the standard
    ProcessNewBlock path (no bypassed validation).
  - CBlockLight (primitives/block.h) + a dedicated job cache (rpc/auxcache.*) backing the above.
  - Light-job cache now clears on chain-tip change, matching the existing createauxblock cache behavior —
    otherwise stale jobs would accumulate indefinitely.
  - New end-to-end test in qa/rpc-tests/createauxblock.py: multi-output coinbase, client-side
    merkle-branch folding, mined AuxPoW, submit, and payout verification.

**Testing**

  python3 qa/rpc-tests/getauxblock.py --srcdir=../../src
  python3 qa/rpc-tests/createauxblock.py --srcdir=../../src


chromatic:
I'm open to the idea. For the implementation:

- is there prior art in other AuxPow/scrypt coin codebases, or is this net new
- would adding default values to `getblocktemplate`, if not provided, achieve the same benefit?

Again, I'm open to the idea but want to probe the potential solution space before committing to an implementation.

ab-emcd:
@chromatic 

Thank you for taking the time to review my PR.
1. The implementation is based on BCH’s lightweight block logic. For AuxPoW coins, we widely use `createauxblocklight` internally.
2. This could be added to `getblocktemplate`, but the benefits of the AuxPoW logic would be lost and shifted to the Stratum side, which I would like to avoid. By “benefits,” I mean the internal cache and transaction-merkle-tree generation. I was trying to avoid moving the entire block over the network, similar to how simple AuxPoW works.

Thanks, Artem B.

